PR #107, opened 2017-09-29
Feature/Replace express exploration project with django production project

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@robinpuga @mackh @dylanmccall Please review. I've tagged Tucker's work in the last commit before this PR so we will be able to loot at or continue work on it as needed.

https://github.com/pg-irc/pathways-backend/releases/tag/MongoExpressExploration

dylanmccall:
Giant commit, so I'll put various thoughts here instead of in a merge review:

 * I think it's tidier if we move the old master branch to a new repository or another branch, then push this code (as a first commit) to an *empty* branch. There's no relationship between this and [MongoExpressExploration](https://github.com/pg-irc/pathways-backend/releases/tag/MongoExpressExploration), so as far as I can tell, basing this on that is only going to pollute the repository and make life more difficult than necessary when we want to look at that old code.
 * I'm not convinced by the addition of a "controllers" directory for each application. May you please elaborate on why that is needed? In Django land, [views *should* serve a similar role to controllers]((https://docs.djangoproject.com/en/1.11/faq/general/#django-appears-to-be-a-mvc-framework-but-you-call-the-controller-the-view-and-the-view-the-template-how-come-you-don-t-use-the-standard-names)). A Django view grabs stuff from models, then sends that stuff to a template to be rendered.
 * I agree about adding a layer around accessing models. However, I'm a little unclear on the scope of ChoiceRepository (and ones like it). Is it supposed to abstract the ORM completely? I think a common convention with Django projects is to have a queries.py that takes care of your database queries (returning [QuerySet objects](https://docs.djangoproject.com/en/1.11/ref/models/querysets/) or [model instance objects](https://docs.djangoproject.com/en/1.11/ref/models/instances/)). Another useful thing is [proxy models](https://docs.djangoproject.com/en/1.11/topics/db/models/#proxy-models) when you don't want to expose an entire table.
 * For settings.py, consider splitting environment-specific settings (like databases, allowed hosts and debug) into a separate local_settings.py.

rasmus-storjohann-PG:
Hi @dylanmccall thanks for your comments, I'll respond in order:

- I used the same repository for the code so that we can keep the connection between the new code and the existing issues. I would also like to keep using the repository name, "pathways-backend", going forward. I tagged the last commit before I removed the mongo project, so that commit is easy to find and new branches can be created from that point for the future.
- I added the controller layer to keep the views simple. I am always trying to apply the single responsibility principle, and the views to me seem to be responsible for interpreting the request, calling business logic (if any), and building the response. Having the view also do the business logic both gives the view too much to do, and makes the view harder to test. Whether "controller" is the right term is an open question. This is also just a first stab at some patterns, I expect that our understanding of this will get better as the system gets more meat on the bones. But with this design, the vote view and vote controllers are both I think quite simple, which is what I was going for.
- If possible, I would like the ORM to be abstracted completely, but if that requires a lot of boiler plate to do, then it's not the way to go. But I like the idea to different repository classes responsible for different queries, generally grouped by data type, i.e. all queries that accept or return a particular model class belong together in the same repository class. One Django pattern that I don't agree with so much is to put too many things in the same file. Having a single queries.py file for the server doesn't scale very well --- if we have more than a dozen queries, that file would very difficult to keep organized.

Looking forward to discussing this further next week.

PR #110, opened 2017-10-20
Feature/add rest support

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@dylanmccall I have followed your suggestions of creating a new repo with the discovery code from Tucker, and fixed up the backend repo to contain just my code. It is much cleaner, and all the issues are still there, so it's looking good. Have a look at this PR with the django REST stuff. Never did figure out how to get URLs in the responses, leaving that for another day, but it will be needed for sure.

dylanmccall:
Cool! That all looks good to me :)

Issue #35, opened 2017-07-31
Establish pattern for multilingual dynamic content


tuckerbuchy:
We would like to ensure our design can support multilingual string localization for our data store in the event that we wish to store the translations in the future.

relevant article from khan academy: http://engineering.khanacademy.org/posts/translation-server.htm

@reyrodrigues will provide a sample data base schema to get this conversation flowing

tuckerbuchy:
@reyrodrigues any updates on this?

tuckerbuchy:
Possible Database schema implementation.
```json
// CURRENT IMPLEMENTATION
// Organization
[
    {
        "id": "1234-1234-1234",
        "name": "Society of People",
        "description": "A place to go and do things"
    }
]


// PROPOSED IMPLEMENTATION
// Organization with translations
[
    {
        "id": "1234-1234-1234",
        "i18n_name": "5667-5566-8887",
        "i18n_description": "5327-5566-8217",
    }
]

// localized/internationalized 
[
    {
        "i18n_id": "5667-5566-8887",
        "locale": "es",
        "value": "Sociedad de Personas"
    },
    {
        "i18n_id": "5327-5566-8217",
        "locale": "es",
        "value": "Un lugar para ir y hacer cosas"
    },
    {
        "i18n_id": "5667-5566-8887",
        "locale": "de",
        "value": "Gesellschaft der Menschen"
    },
    {
        "i18n_id": "5327-5566-8217",
        "locale": "de",
        "value": "Ein Ort, um zu gehen und Dinge zu tun"
    },
]
```

rasmus-storjohann-PG:
AC: Have an understanding of what is required, either in terms of libraries to use or database schemas to build.

PR #117, opened 2017-11-01
Issue #35 Add dependency on django-parler

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@dylanmccall please review

PR #122, opened 2017-11-03
Feature/issue 118 create thin service povider data model

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@dylanmccall for review. My estimates are too high, this stuff goes a lot faster than I think it will. We'll see how long that lasts...

service_providers/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+from parler.models import TranslatableModel, TranslatedFields
+
+class ServiceProvider(TranslatableModel):
+    name = models.CharField(max_length=200)
+    latitude = models.DecimalField(max_digits=9, decimal_places=6)
+    longitude = models.DecimalField(max_digits=9, decimal_places=6)
+    translations = TranslatedFields(
+        description=models.TextField(null=True)
```
dylanmccall: `null=True` probably isn't quite what you're aiming for here. That's saying that the database should allow NULL as a possible value for the description field, but it doesn't affect model validation. (As far as built-in forms and validators are concerned, you still must enter a description). What you want here is probably a combo of `blank=True, null=True`. The blank attribute says that the field is allowed to be blank. Folks generally like to skip `null=True` altogether, since it has the same meaning as an empty string, but this is up to our own discretion (as long as we are consistent).

service_providers/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+from parler.models import TranslatableModel, TranslatedFields
+
+class ServiceProvider(TranslatableModel):
+    name = models.CharField(max_length=200)
+    latitude = models.DecimalField(max_digits=9, decimal_places=6)
+    longitude = models.DecimalField(max_digits=9, decimal_places=6)
```
dylanmccall: It may be worth looking at [GeoDjango](https://docs.djangoproject.com/en/1.11/ref/contrib/gis/) (and its Point field, in this case) for this project. Or some other system that provides spatial indexes. May be out of scope for this particular change, though.

service_providers/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class ServiceProvidersConfig(AppConfig):
+    name = 'service_providers'
```
dylanmccall: While we're here, it's helpful to add a localized verbose name for the application as well:

    verbose_name = _("Service providers")

Otherwise Django just shows its own guess at the name in the admin panel.

rasmus-storjohann-PG: Fixed

service_providers/models.py
rasmus-storjohann-PG: Thanks for the link. I will put off this decision for now, as long as what you get over the wire will let you show the service providers on the map, that's all we need.

service_providers/models.py
rasmus-storjohann-PG: So given that this is a text field, which is never NULL in django, but "" in stead, the correct thing to use here would be `blank=True`. Fixed

Issue #108, opened 2017-10-05
Describe a data model for the backend


dylanmccall:
Complete by October 10, 2017.

dylanmccall:
Here's my initial stab at how this should work:

https://gist.github.com/dylanmccall/8b7a18fd4a68afb1606676fdc06844a3

Needs a pretty diagram.

dylanmccall:
And here's a pretty diagram. ![pathways data model - backend relationship diagram](https://user-images.githubusercontent.com/132063/31416484-6151010e-adde-11e7-824f-ae419b21de43.png).

PR #123, opened 2017-11-07
Feature/issue 119 create api for service providers

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@dylanmccall for review

mysite/urls.py
```diff
@@ -4,6 +4,7 @@
 # pylint: disable=invalid-name
 urlpatterns = [
     url(r'^v1/forms/polls/', include('polls.web.urls')),
-    url(r'^v1/', include('polls.web.api')),
+    url(r'^v1/polls/', include('polls.web.api')),
+    url(r'^v1/service-providers', include('service_providers.urls')),
```
dylanmccall: I'm a little skeptical of the service_providers application being only for the one model. In #108 we split our models between three different applications, where each one provides a few different models. Externally it can look identical, but obviously we'd need to structure our url.py a bit differently. Instead of `url(r'^v1/service-providers', include('service_providers.urls'))`, we'd want `url(r'^v1/', include('service_providers.urls'))`, and then service_providers/urls.py would provide mappings for 'services-providers/…', etc.

dylanmccall:
Probably out of scope here, but, incidentally, here's some stuff about versioning with django-rest-framework so we aren't just hard-coding v1 in our urls:

http://www.django-rest-framework.org/api-guide/versioning/

service_providers/urls.py
```diff
@@ -0,0 +1,11 @@
...
+from django.conf.urls import url, include
+from service_providers import viewsets
+
+def build_urlpatterns():
+    router = routers.DefaultRouter()
+    router.register(r'^', viewsets.ServiceProviderViewSet)
+    return [url(r'^', include(router.urls))]
+
+# pylint: disable=invalid-name
+urlpatterns = build_urlpatterns()
```
dylanmccall: Looking at it again, this URL structure seems really weird to me. You'll notice that a request for _/v1/service-providers_ returns a default API root listing, while _/v1/service-providers/_ returns the service providers view set. That happens because router.register expects an actual prefix for its set of roots. DefaultRouter already provides its own root URL mapping. Is there a reason this can't be something more like this?

    router.register(r'^service-providers', viewsets.ServiceProviderViewSet)
    
    # pylint: disable=invalid-name
    urlpatterns = router.urls

mysite/urls.py
rasmus-storjohann-PG: But if we have `url(r'^v1/', include('service_providers.urls'))`, what do we do with pathways.urls, it would also be under `v1`

service_providers/urls.py
rasmus-storjohann-PG: I was struggling with this, so I do know there is a problem. We just need to be able to hang different applications off of `v1/`, so somehow the element following `v1/` has to be in the top level urls file.

mysite/urls.py
rasmus-storjohann-PG: Oh I see there's a gist too, not just the UML. I will look at this.

service_providers/urls.py
dylanmccall: There are a few different approaches here. One is we can do this…

    url(r'^v1/', include('someapp.urls')),
    url(r'^v1/', include('anotherapp.urls')), 

It works in a pinch, although that's a bit ugly since we are assuming the first app doesn't try to handle any URLs we want to pass in to the second.

Another approach is you add a third app which is specifically responsible for the URLs for your API. So it creates a single DefaultRouter, loads the viewsets from the other apps, and registers them. I've seen some folks doing basically the same thing by putting more logic in their root urls.py, as well, if you prefer.

rasmus-storjohann-PG:
@dylanmccall please have a look at the last commit, I think it makes some good improvements.

dylanmccall:
Thanks! That looks better to me :)

PR #124, opened 2017-11-08
Add django-cors-headers to allow requests from other domains

dylanmccall:
Browsers are very persnickety about cross-origin HTTP requests. When our web app tries to do a fetch from the API, there's a preflight OPTIONS request, and we need our app to say "yup, this is okay". This adds a module which does that for us, and some very open-ended initial configuration.

### DIFF
--- a/mysite/settings.py
+++ b/mysite/settings.py
@@ -15,6 +15,8 @@ DEBUG = True
 
 ALLOWED_HOSTS = []
 
+CORS_ORIGIN_ALLOW_ALL = True
+
 # Application definition
 
 INSTALLED_APPS = [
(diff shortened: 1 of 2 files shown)


mysite/settings.py
```diff
@@ -148,6 +150,12 @@
 USE_TZ = True
 
 
+# CORS headers
+# https://github.com/ottoyiu/django-cors-headers
```
rasmus-storjohann-PG: Remove comments, google will point the way if we ever need to look this up

requirements.txt
```diff
@@ -1,7 +1,9 @@
+## The following requirements were added by pip freeze:
```
rasmus-storjohann-PG: Remove comment

dylanmccall: That comment is just generated by pip freeze. Didn't realize it wasn't there originally, but it isn't really necessary to remove it. FYI, it's a good idea to update requirements.txt like this:

    pip freeze -r requirements.txt > requirements.txt

That ensures pip will maintain the existing ordering, comments, etc. in the file.

PR #135, opened 2017-11-17
Feature/issue 120 parse bc211 data from xml

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@dylanmccall for review.

bc211/importer.py
```diff
@@ -0,0 +1,29 @@
+from bc211 import parser
+from service_providers import models
+from django.utils import translation
+
+def read_records_from_file(path):
```
dylanmccall: This would be a little more helpful if it accepted a file object. Otherwise we are assuming that "file" is a file on the disk, as opposed to [StringIO](https://docs.python.org/3/library/io.html) or [sys.stdin](https://docs.python.org/3/library/sys.html#sys.stdin), for instance.

Either way, I find it useful to refer to file objects with "_file", and file paths with "_path". Otherwise this stuff gets pretty muddled.

bc211/models.py
```diff
@@ -0,0 +1,12 @@
+from bc211 import validators
+
+class SpatialLocation:
```
dylanmccall: It's a bit confusing that this file is called models.py when it doesn't include any models. Are these relevant outside of parser.py?

service_providers/management/commands/import_bc211_data.py
```diff
@@ -0,0 +1,20 @@
...
+from bc211.importer import read_records_from_file, save_records_to_database
+
+# invoke as follows:
+# python manage.py import_bc211_data path/to/bc211.xml
+
+class Command(BaseCommand):
+    help = 'Import BC-211 data from XML file'
+
+    def add_arguments(self, parser):
+        parser.add_argument('file', metavar='file', help='Path to XML file containing BC-211 data')
```
dylanmccall: Continuing on that other comment about files, here's something really cool you can do with argparse:

    import argparse
    parser.add_argument('file', type=argparse.FileType('r'), metavar='file', help='Path to XML file containing BC-211 data')

By setting a type, you get validation for free, and options['file'] will be an actual file object.

That's documented over here: https://docs.python.org/3/library/argparse.html#argparse.FileType.

bc211/tests/test_importer.py
```diff
@@ -0,0 +1,41 @@
...
+from decimal import Decimal
+from django.test import TestCase
+from bc211.importer import read_records_from_file, save_records_to_database
+from service_providers import models
+
+INCLUDE_SLOW_TESTS = False
+
+ONE_AGENCY_FIXTURE = 'bc211/data/BC211_data_one_agency.xml'
+MULTI_AGENCY_FIXTURE = 'bc211/data/BC211_data_excerpt.xml'
+ALL_AGENCIES_FIXTURE = 'bc211/data/AIRSXML_2252_Export_20170109050136_BC_211.xml'
```
dylanmccall: Not a big deal, but these relative paths bug me. You could get a base directory for the project by importing it from mysite.settings. You can also get the directory for this test file with `os.path.dirname(__file__)` and build a relative path from there.

bc211/parser.py
```diff
@@ -0,0 +1,26 @@
+import xml.etree.ElementTree as etree
+from bc211 import models
+
+def parse_all_service_providers(xml_data_as_string):
+    root_xml = etree.fromstring(xml_data_as_string)
+    all_providers_xml = root_xml.findall('Agency')
+    return list(map(parse_one_service_provider, all_providers_xml))
```
dylanmccall: The [map](https://docs.python.org/3/library/functions.html#map) function returns an iterator, so we aren't required to keep everything in memory. That's great, because service providers live in a 34.4MB XML file. Unfortunately, list() takes that and throws it out: now we have to iterate the entire map right there and stick everything in memory along with some other stuff related to dynamic lists.

It looks like the only list-related thing we are doing with this is `len(records)` in import_bc211_data.py. I'd suggest changing this function to return an iterator (so, don't wrap *map* in *list*), then find ways to count our results without using len(). You may like the [enumerate](https://docs.python.org/3/library/functions.html#enumerate) function for future reference, although in this case just incrementing a count variable inside importer.save_records is probably more appropriate.

bc211/importer.py
rasmus-storjohann-PG: Can you explain why it's muddled otherwise?

rasmus-storjohann-PG:
The rest of your comments make a lot of sense, I will work on it tomorrow

rasmus-storjohann-PG:
Hey @dylanmccall thanks for your comments, they were very helpful. I have made changes, please have a look and merge if you think it's good to go. I have not addressed all of your comments, we can talk about the ones I have not.

dylanmccall:
That looks good! Glad it helped :)

PR #142, opened 2017-11-30
Add an application to export and import PO files with site content

### LINKED ISSUE #132: Export and import site content PO file for translators
* [x] Add a management command, similar to [makemessages](https://docs.djangoproject.com/en/1.11/ref/django-admin/#makemessages), that creates a single message file for all TranslatableModel records. It would be a useful optimization to update an existing message file with changes, similar to msgmerge. The command should accept a domain (message file name) as a parameter, defaulting to "django_content".
 * [x] Add a management command that acts as an inverse of makemessages. It should accept a message file, or a set of message files, and use them to update translations belonging to TranslatableModel records.
 * [ ] Add tests for the parler_po application.

The [polib](http://polib.readthedocs.io/en/latest/quickstart.html) library may be helpful here. It provides tools for working with message files.

AC: Using a command line interface, I can export all strings from the database that require translation and produce a PO file as per the API requirements of Transifex and similar translation services. Using a command line interface, I can also import strings from a PO file to populate language-specific columns in the database.

dylanmccall:
This adds some management commands under the parler_po application.

List available translatable models:

    ./manage.py content_translation_list_models

Create a POT file for a particular translatable model:

    ./manage.py content_translation_export organizations.organization > organization.pot

Create a PO file for a particular translatable model in French:

    ./manage.py content_translation_export organizations.organization --language=fr > fr/organization.po

The resulting files can be opened in an application such as [GTranslator](https://wiki.gnome.org/Apps/Gtranslator) or [Poedit](https://poedit.net/).

After you have updated your translations, you can import them again:

    ./manage.py content_translation_import fr/translation.po

Closes #132


rasmus-storjohann-PG:
General comments:
* First of all, this is good work, I understand that this has taken you longer than we initially anticipated. The test coverage is still not where it needs to be, and I think that once you start writing tests as you go, you will be able to make faster progress. I will ask you to complete the tests for the code in this PR before it gets merged. I know that this will feel like wasted effort and very slow, but I can assure you that it will make us faster in the long run. Having tests in place will make it faster to understand what the tests do, and much much faster to make changes to the code, since each time we make a change, we can spend 2 seconds verifying that we didn't break anything, rather than probably 10 minutes of repetitive, boring, manual testing that we both know wouldn't be done and bugs will creep in, slowing us down further. Let's talk about what tests you need to write and a little bit about how to write them.
* The name `parler_po` can be improved. Remember I mentioned the naming problem we had at EA with naming a lot of the code using the codename of the product before it was launched, we ended up with code that was littered with names that no longer had any meaning. Naming our code according to third party libraries is similar, if we one day switch from parler to some other library, or the next new translation system that doesn't use po and pot files, then this name will mean nothing. I suggest the name should be `translation`.
* Let's decide on whether we use single or double quotes for strings. So far I have consistently used single quotes since they are less visually noisy.
* Let's not localize logging and error messages. For error messages, we should create specific exception classes for errors and have the top level exception handler contain any localization we might need for that.
* Logging: Have a look at existing code and follow the pattern of using the python logging mechanism rather than printing straight to standard out.
* Abbreviations in code tends to hurt readability for little gain, I will highlight where I think renames can help.
* I'm happy that you are putting the issue numbers in the commit messages, can I ask that you go the extra mile and adhere to the established pattern in the repository and use `Issue #xxx` in the commit messages.

parler_po/management/commands/parler_po_export.py
```diff
@@ -0,0 +1,123 @@
...
+                    language_code__in=locales
+                )
+
+            for translation in translations_query:
+                po_entries = self._translation_po_entries(
+                    translatable,
+                    base_translation,
+                    translation=translation
+                )
+                yield (translation.language_code, po_entries)
```
rasmus-storjohann-PG: My core principle in writing software is the single responsibility principle. There are many functions in this PR that do more than one thing, but this one stands out with having multiple yield in one function. I don't even know how that works, and I would strongly prefer to never have to know.

parler_po/util.py
```diff
@@ -0,0 +1,74 @@
...
+    po_file.metadata['Content-Transfer-Encoding'] = '8bit'
+
+    return po_file
+
+def get_pot_path(output_dir, model):
+    content_type = ContentType.objects.get_for_model(model)
+    domain = content_type_id(content_type)
+    pot_name = '{}.pot'.format(domain)
+    os.makedirs(output_dir, exist_ok=True)
+    return os.path.join(output_dir, pot_name)
```
rasmus-storjohann-PG: I think the name `get_foo` implies that the function returns me something that the callee already has. When the function is actually creating or building something, `create_foo` or `build_foo` will be clearer.

parler_po/util.py
```diff
@@ -0,0 +1,74 @@
...
+PARLER_PO_BASE_LANGUAGE = getattr(settings, 'PARLER_PO_BASE_LANGUAGE', 'en')
+PARLER_PO_CONTACT = getattr(settings, 'PARLER_PO_CONTACT', None)
+
+def get_base_translation(translatable):
+    if translatable.has_translation(PARLER_PO_BASE_LANGUAGE):
+        return translatable.get_translation(PARLER_PO_BASE_LANGUAGE)
+    else:
+        return None
+
+def new_pot_file():
```
rasmus-storjohann-PG: The function prefix `new` can also be replaced with `create` or `build`. While not an issue in Python, recovering C++ and Java programmers will have strong connotations with `new` that are not really lining up with what this function does.

parler_po/management/commands/parler_po_import.py
```diff
@@ -0,0 +1,174 @@
...
+class ImportProgress(object):
+    PROGRESS_ERROR = 0
+    PROGRESS_NEW = 1
+    PROGRESS_SKIP = 2
+
+    def __init__(self):
+        self._counts = defaultdict(lambda: defaultdict(int))
+
+    def add_error(self, key):
+        self._counts[key][None] += 1
```
rasmus-storjohann-PG: This can be made clearer if you do `self._counts[key][self.PROGRESS_TOTAL] += 1`

parler_po/translation_entry.py
```diff
@@ -0,0 +1,89 @@
...
+            msgid=self.msgid,
+            msgstr=self.msgstr,
+            occurrences=[(self.instance_field_id, None)]
+        )
+
+    def get_base_translation(self):
+        return get_base_translation(self.instance)
+
+    def get_translation(self, language_code):
+        base_translation = self.get_base_translation()
```
rasmus-storjohann-PG: My head hurts a bit from navigating multiple functions with identical names here, especially with Python's weaker support for overloading. Renaming some functions here will make it easier to follow.

parler_po/management/commands/parler_po_export.py
dylanmccall: Multiple yields are actually really useful for flattening trees, or for presenting tree-like logic in a flat way. It works exactly the same as having a yield inside a loop. Agreed, though, that function is ugly as sin. Shall rewrite it.

parler_po/management/commands/parler_po_export.py
```diff
@@ -0,0 +1,123 @@
...
+        else:
+            msg = _("Skipping \"{record}\": Missing base translation").format(
+                record=translatable
+            )
+            self.stderr.write(self.style.WARNING(msg))
```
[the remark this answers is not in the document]
dylanmccall: Wasn't sure which pattern we were using here. I'll switch to the latter. I tend to prefer the first way because it discourages functions with too many nested conditions (usually a bad sign), but it's definitely a bit excessive here.

parler_po/management/commands/parler_po_import.py
dylanmccall: Will do! Thanks.

parler_po/translation_entry.py
dylanmccall: Yeah, this one should probably be get_translation_for_language, which would already help…

parler_po/util.py
dylanmccall: Good catch. Thanks!

parler_po/util.py
dylanmccall: Trouble with "create" is it tells a Django developer that it's going to make a new object and then commit it to the database (or to the filesystem, in this case), which isn't what this function should be doing. Thinking about it, this might be better as a helpful POFile subclass.

dylanmccall:
Thanks for the comments! I'll address these all soon.

 * I like to use single quotes for strings intended for computers (like hard-coded field names); and double quotes for strings intended for humans (which are usually strings we would translate, like help text). I find that convention in quite a few places. If you'd prefer single quotes everywhere that's just fine. We should stick it in a file about coding conventions.
 * Will remove localization for errors and log messages. As long as we're consistent one way or another I'm happy.
 * I'm a little hesitant to use logging for management commands because we usually want some direct control in them with regards to what goes to stdout and what goes to stderr; as well as text formatting, which we aren't concerned about when we're logging server activity. For example, it is helpful for an administrator if we print import progress in a single line (with a \r ending), rather than hammering the terminal with newlines, but that isn't something we can really do with the logging library. Might be good to mix and match, though, at least for output that we aren't as picky about. I'll take a look.
 * I assume you don't like my `msg = …` stuff. Those *are* pretty bad, aren't they? Should probably just have some helper methods that print a string to either logging or stdout with the correct style.
 * Remind me to rebase the branch with updated commit summaries :)

rasmus-storjohann-PG:
Regarding using logging with control over newline/carriage return characters, have a look at this https://docs.python.org/3/library/logging.handlers.html#streamhandler and see if it can solve that problem

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+        app_label = 'parlerpo'
+        managed = False
+
+    static = models.CharField(max_length=100)
+
+    translations = TranslatedFields(
+        foo=models.CharField(max_length=100)
+    )
+
+    # Crude hack to mock TranslatableModel database operations
```
rasmus-storjohann-PG: If you have a hard time mocking a third party interface that you do not control, then wrap that interface in one that you do control and can work with more easily.

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+class CreateTranslationEntryTests(TestCase):
+    def test_create__translatable(self):
+        translatable = TestTranslatable()
+        translatable.id = 1
+        translation_entry = TranslationEntry(translatable, 'foo', 'bar', 'baz')
+
+        self.assertEquals(translation_entry.instance, translatable)
+        self.assertEquals(translation_entry.field_id, 'foo')
+        self.assertEquals(translation_entry.msgid, 'bar')
+        self.assertEquals(translation_entry.msgstr, 'baz')
```
rasmus-storjohann-PG: Avoid abbreviations. Every time we read this code, we have to map from msgid to message id and msgstr to message string. Your future self will be grateful if you take that load off and just call them `message_id` and `message_string`. Also, it is a very good general rule to try to have no more than one assert per test --- this ties in with how tests are named, multiple asserts means that the test is not focused and therefore there is no way to name them to reflect what they are testing for.

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+
+        self.assertEquals(str(translation_entry), 'parlerpo.testtranslatable@foo@1')
+
+    def test_create__not_translatable(self):
+        not_translatable = TestNotTranslatable()
+
+        create_entry_fn = lambda: TranslationEntry(not_translatable, 'foo', 'bar', 'baz')
+        self.assertRaises(TypeError, create_entry_fn)
```
rasmus-storjohann-PG: Rename to `create_from_non_translatable_model_throws` There is a better way to test for raised exceptions using `with`, see e.g. https://github.com/pg-irc/pathways-backend/blob/dcf84c141864ba880eeaa8b15c2ff4d4404d1d82/bc211/tests/test_dtos.py#L6-L7

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+    def create_translation(self, language_code):
+        self.get_translation(language_code)
+
+class TestNotTranslatable(models.Model):
+    class Meta:
+        app_label = 'parlerpo'
+        managed = False
+
+class CreateTranslationEntryTests(TestCase):
+    def test_create__translatable(self):
```
rasmus-storjohann-PG: Why the double underscore? It's a very subtle thing so if it means something important it will be missed

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+
+        self.assertEquals(str(translation_entry), 'parlerpo.testtranslatable@foo@1')
+
+    def test_create__not_translatable(self):
+        not_translatable = TestNotTranslatable()
+
+        create_entry_fn = lambda: TranslationEntry(not_translatable, 'foo', 'bar', 'baz')
+        self.assertRaises(TypeError, create_entry_fn)
+
+    def test_create__static_field(self):
```
rasmus-storjohann-PG: This test does not explain what is being tested here. There are lots of resources on how to name tests, I'm not wedded to any particular scheme.

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+        translatable = TestTranslatable()
+
+        create_entry_fn = lambda: TranslationEntry(translatable, 'static', 'bar', 'baz')
+        self.assertRaises(ValueError, create_entry_fn)
+
+    def test_create__invalid_field(self):
```
rasmus-storjohann-PG: Rename to `create_with_nonexisted_field_throws`

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+            translation_entry = TranslationEntry.from_po_entry(
+                po_entry,
+                po_entry.occurrences[0]
+            )
+            parse_instance_field_id.assert_called_once_with('parlerpo.testtranslatable@foo@1')
+
+        self.assertEquals(translation_entry.instance, translatable)
+        self.assertEquals(translation_entry.field_id, 'foo')
+        self.assertEquals(translation_entry.msgid, 'bar')
+        self.assertEquals(translation_entry.msgstr, 'baz')
```
rasmus-storjohann-PG: This test should be broken apart, both because of the many asserts and the amount of set up code. Only one of these resulting tests will use the mock, so the code will become much simpler.

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+        create_entry_fn = lambda: TranslationEntry(translatable, 'not_a_field', 'bar', 'baz')
+        self.assertRaises(ValueError, create_entry_fn)
+
+    def test_create_from_po_entry(self):
+        translatable = TestTranslatable()
+        translatable.id = 1
+
+        po_entry = polib.POEntry()
+        po_entry.occurrences = [
+            ('parlerpo.testtranslatable@foo@1', None)
```
rasmus-storjohann-PG: I don't understand what `None` means here. One way to make this clear is to use a local variable
```
some_explanatory_name = None
po_entry.occurrences = [
             ('parlerpo.testtranslatable@foo@1', some_explanatory_name)
```
You can also very easily use the 'baz' and 'bar' strings to explain what they are.

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+                po_entry.occurrences[0]
+            )
+            parse_instance_field_id.assert_called_once_with('parlerpo.testtranslatable@foo@1')
+
+        self.assertEquals(translation_entry.instance, translatable)
+        self.assertEquals(translation_entry.field_id, 'foo')
+        self.assertEquals(translation_entry.msgid, 'bar')
+        self.assertEquals(translation_entry.msgstr, 'baz')
+
+    def test_create_from_po_entry__not_translatable(self):
```
rasmus-storjohann-PG: This test also needs to be called something so that I know what is supposed to happen in this usecase

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+            parse_instance_field_id.return_value = (translatable, 'foo')
+            create_entry_fn = lambda: TranslationEntry.from_po_entry(
+                po_entry,
+                po_entry.occurrences[0]
+            )
+            self.assertRaises(TypeError, create_entry_fn)
+
+    def test_create_from_translation(self):
+        translatable = TestTranslatable()
+        translatable.id = 1
```
rasmus-storjohann-PG: Pull repeated code out into setUp()

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+
+        base_translation = translatable.get_translation('en')
+        base_translation.foo = 'bar'
```
rasmus-storjohann-PG: So base translation is the data in the source language? Since the defining characteristic of this data is that it is not translated, calling it a translation is not ideal. I think it should be called something with 'source', maybe `source_data` or `source_strings`, as opposed to `translated_strings`.

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+        self.assertEquals(translation_entry.instance, translatable)
+        self.assertEquals(translation_entry.field_id, 'foo')
+        self.assertEquals(translation_entry.msgid, 'bar')
+        self.assertEquals(translation_entry.msgstr, 'baz')
+
+class ValidTranslationEntryTests(TestCase):
+    TRANSLATION_FIELD = 'foo'
+    TRANSLATION_MSGID = 'bar'
+    TRANSLATION_MSGSTR = 'baz'
+    TRANSLATABLE_ID = 1
```
rasmus-storjohann-PG: What's the convention for all upper case here?

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+        ])
+
+    def test_as_po_entry__no_msgid(self):
+        self.translation_entry.msgid = ''
+
+        as_translation_fn = lambda: self.translation_entry.as_po_entry()
+
+        self.assertRaises(ValueError, as_translation_fn)
+
+    def test_as_translation(self):
```
rasmus-storjohann-PG: Rename to `converting_to_translation_sets_master_to_translatable_model()` and similar for the other, broken out tests.

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+class ValidTranslationEntryTests(TestCase):
+    TRANSLATION_FIELD = 'foo'
+    TRANSLATION_MSGID = 'bar'
+    TRANSLATION_MSGSTR = 'baz'
+    TRANSLATABLE_ID = 1
+
+    def setUp(self):
+        self.translatable = TestTranslatable()
+        self.translatable.id = self.TRANSLATABLE_ID
+        self.translatable.foo = self.TRANSLATION_MSGID
```
rasmus-storjohann-PG: This can be made much more readable by changing `foo` to something meaningful here.

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+
+    def test_as_po_entry(self):
+        po_entry = self.translation_entry.as_po_entry()
+
+        self.assertEquals(po_entry.msgid, self.TRANSLATION_MSGID)
+        self.assertEquals(po_entry.msgstr, self.TRANSLATION_MSGSTR)
+        self.assertEquals(po_entry.occurrences, [
+            ('parlerpo.testtranslatable@{}@{}'.format(self.TRANSLATION_FIELD, self.TRANSLATABLE_ID), None)
+        ])
```
rasmus-storjohann-PG: This last assert is actually testing a whole bunch things, and it is likely more specific than it needs to be. What are the things you really care about here? You care that the model name, field and and id all appear in the computed strings, so that is three tests, each of which can just use a string contains assertion. And you care that occurrences contains exactly one entry, that's one more test. Do you really care about the formatting with the '@'s? If so, that is one more test, but I am guessing that this is not really a requirement. Leaving that test out have the benefits that you can change that formatting without breaking the tests, and the tests are actually more expressive. The tests should not test unimportant details, that way they actually tell us what about the implementation matters and (by omission) what does not.

parler_po/tests/test_translation_entry.py
```diff
@@ -0,0 +1,179 @@
...
+        self.assertRaises(ValueError, as_translation_fn)
+
+    def test_as_translation(self):
+        translation = self.translation_entry.as_translation('fr')
+
+        self.assertEquals(translation.master, self.translatable)
+        self.assertEquals(translation.foo, self.TRANSLATION_MSGSTR)
+        self.assertTrue(translation.is_modified)
+
+    def test_as_translation__base_translation(self):
```
rasmus-storjohann-PG: A better test name will make it clear what is going on here, `converting_source_entry_to_translation_returns_none`. Is this the correct behaviour though? Never return NULL is a good rule in C++, and I think it applies here as well. I think this call should throw.

parler_po/translation_entry.py
```diff
@@ -0,0 +1,109 @@
+from parler.models import TranslatableModel
+import logging
+import polib
+
+from parler_po.util import (
+    get_base_translation,
+    build_instance_field_id,
+    parse_instance_field_id
+)
```
rasmus-storjohann-PG: I think `from parler_po import util` is better, or `from parler_po import util as po_utils`

parler_po/management/commands/parler_po_import.py
```diff
@@ -0,0 +1,171 @@
+from collections import defaultdict
+from django.core.management.base import BaseCommand, CommandError
+from django.db import models
+from django.utils.translation import ugettext as _
+from django.utils.translation import ungettext as __
```
rasmus-storjohann-PG: I meant to mention this earlier, having a function called `__` as well as `_` is asking for future confusion and bugs. I would prefer to just use `ungettext()` in the few places where it is called, it will make it clear what is being called and will highlight the few placed where we need to do something unusual for localization.

parler_po/management/commands/parler_po_import.py
```diff
@@ -0,0 +1,171 @@
...
+                    msg = "Skipping \"{occurrence}\": {error}".format(
+                        occurrence=":".join(n for n in occurrence if n),
+                        error=error
+                    )
+                    self.stderr.write(self.style.WARNING(msg))
+                else:
+                    yield translation_entry
+
+    def _import_translation_entry(self, translation_entry, language_code, import_progress):
+        import_group = self._get_import_group(translation_entry, language_code)
```
rasmus-storjohann-PG: So about the _ prefix on private functions. I understand that this is a convention in python, and I guess that is OK, even though they are not actually private. To have proper private functions, it is better to make a submodule that I would always see called `details` and put all "private" functions there.

parler_po/translation_entry.py
```diff
@@ -0,0 +1,109 @@
...
+
+class TranslationEntry(object):
+    def __init__(self, instance, field_id, msgid, msgstr):
```
rasmus-storjohann-PG: Naming: instance is a very generic term, can we have a name that has more meaning, maybe row or record to tie this to the fact that this is a database row? Open for suggestions. Fix msg to message.

parler_po/translation_entry.py
```diff
@@ -0,0 +1,109 @@
...
+        (instance_field_id, _lineno) = occurrence
+
+        instance, field_id = parse_instance_field_id(instance_field_id)
+
+        return cls(
+            instance,
+            field_id,
+            msgid=po_entry.msgid,
+            msgstr=po_entry.msgstr
+        )
```
rasmus-storjohann-PG: Since you can't call it `class` I would be happy with `the_class` or `type`. 
It's a matter of taste whether this function should be a class method or a free function, I would do the latter but I'm fine either way.

parler_po/translation_entry.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @classmethod
+    def from_translation(cls, translation, field_id):
+        base_translation = get_base_translation(translation.master)
+
+        if base_translation:
+            msgid = getattr(base_translation, field_id, '')
+        else:
+            msgid = ''
```
rasmus-storjohann-PG: These five lines will make a really nice function

parler_po/translation_entry.py
```diff
@@ -0,0 +1,109 @@
...
+            msgid=self.msgid,
+            msgstr=self.msgstr,
+            occurrences=[(self._instance_field_id, None)]
+        )
+
+    def as_translation(self, language_code):
+        base_translation = get_base_translation(self.instance)
+
+        if self.msgid != getattr(base_translation, self.field_id):
+            raise ValueError("Invalid msgid")
```
rasmus-storjohann-PG: Do you have tests covering this error condition? I can't remember.

parler_po/translation_entry.py
```diff
@@ -0,0 +1,109 @@
...
+        else:
+            translation = None
+
+    # TODO: If msgstr is blank, should we fall back to the base language? Or
+    #       should we expect translators to do that in their PO files?
+
+    if translation:
+        setattr(translation, field_id, msgstr)
+
+    return translation
```
rasmus-storjohann-PG: Function does too much.

parler_po/tests/test_translation_entry.py
dylanmccall: Constants tend to be defined in all capital capital letters with underscores. (See PEP-8 on naming constants: https://www.python.org/dev/peps/pep-0008/#constants).

parler_po/tests/test_translation_entry.py
dylanmccall: That's really helpful, thanks! I didn't realize assertRaises could be used as a context manager, but that makes lots of sense :)

parler_po/tests/test_translation_entry.py
dylanmccall: Kind of CSS-ism: some-thing--some-modifier. I'll swap that with more explicit names like you suggested.

dylanmccall:
Okay, if you can give this another look now, it would be much appreciated :)

The tests are unfortunately a bit erratic in terms of what they cover, but should stretch this thing in some helpful ways.

parler_po/exceptions.py
```diff
@@ -0,0 +1,56 @@
...
+class ModelNotTranslatableError(ParlerPOError):
+    def __init__(self, model):
+        self.model = model
+        message = "{} is not a TranslatableModel.".format(model.__name__)
+        super().__init__(message)
+
+class FieldNotTranslatableError(ParlerPOError):
+    def __init__(self, field_id):
+        self.field_id = field_id
+        message = "{} is not a translatable field.".format(field_id)
```
rasmus-storjohann-PG: We said single quotes, right?

parler_po/exceptions.py
```diff
@@ -0,0 +1,56 @@
...
+    pass
+
+class MasterInstanceLookupError(ParlerPOError):
+    def __init__(self, message):
+        super().__init__(message)
+
+class ModelNotTranslatableError(ParlerPOError):
+    def __init__(self, model):
+        self.model = model
+        message = "{} is not a TranslatableModel.".format(model.__name__)
```
rasmus-storjohann-PG: When handling errors, it's good to keep an eye out for things that can throw, since throwing while handling an exception is really bad news. I would look for a way to construct this exception in a way that we know can never throw.

parler_po/import_progress.py
```diff
@@ -0,0 +1,63 @@
...
+            numbers_list.append(
+                str(total_count)
+            )
+
+            new_count = counts[self.PROGRESS_NEW]
+            if new_count:
+                numbers_list.append(
+                    ungettext("({} new)", "({} new)", new_count).format(
+                        new_count
+                    )
```
rasmus-storjohann-PG: It's not clear what this is doing, a local variable or two with good names will fix that

parler_po/management/commands/parler_po_export.py
```diff
@@ -0,0 +1,106 @@
...
+from parler.models import TranslatableModel
+import itertools
+
+from parler_po.argparse_path import ArgparsePathType
+from parler_po.exceptions import ParlerPOError
+from parler_po.po_file import create_po_file, create_pot_file
+from parler_po.queries import all_translatable_models, get_base_translation
+from parler_po.translatable_string import TranslatableString
+
+class Command(BaseCommand):
```
rasmus-storjohann-PG: You don't need to change this in this PR, but by the Single Responsibility Principle, this class should deal with being a command only, and all business logic should be delegated to a separate class. That class would also be much easier to test, I'm guessing...

parler_po/tests/helpers.py
```diff
@@ -0,0 +1,29 @@
+from organizations.models import Organization
+
+BASE_LANGUAGE = 'en'
```
rasmus-storjohann-PG: You don't need to change this now, but: The benefit of the builder pattern is that you can define defaults in the constructor and easily use other values instead by calling withFoo() functions. Having a hard-coded default as a constant is at cross purposes with the builder pattern.

parler_po/tests/helpers.py
```diff
@@ -0,0 +1,29 @@
...
+            self.instance.create_translation(language_code)
+
+        translation = self.instance.get_translation(language_code)
+
+        for field_name, value in fields.items():
+            setattr(translation, field_name, value)
+
+        translation.save()
+
+        return self
```
rasmus-storjohann-PG: I would definitely put all of this in the build() function, so that the with() function does not have side effects

parler_po/tests/test_field_ids.py
```diff
@@ -0,0 +1,101 @@
...
+    ContentTypeDoesNotExistError,
+    InvalidContentTypeIDError,
+    InvalidInstanceFieldIDError,
+    ModelInstanceDoesNotExistError
+)
+from parler_po.field_ids import (
+    build_instance_field_id,
+    parse_instance_field_id,
+    _build_content_type_id,
+    _parse_content_type_id
```
rasmus-storjohann-PG: Are you depending on internals of a third party library here? That will bite us

parler_po/tests/test_parler_po_export.py
```diff
@@ -0,0 +1,210 @@
...
+            OrganizationBuilder(id='two').with_base_translation(
+                description='organization_two_translation_description_msgid'
+            ).with_translation(
+                'fr', description='organization_two_translation_description_msgstr_fr'
+            ).build(),
+            OrganizationBuilder(id='three').with_base_translation(
+                name='organization_three_translation_name_msgid',
+                description='organization_three_translation_description_msgid'
+            ).build()
+        ]
```
rasmus-storjohann-PG: This can be made so much more readable so easily. the builder clearly needs a `with_translation_name()` setter, which solves the complicated `with_translation()` calls. Then three local variables, one for each organization, will solve the 15 function calls to build one small array problem.

rasmus-storjohann-PG:
> The tests are unfortunately a bit erratic in terms of what they cover, but should stretch this thing in some helpful ways. 

Writing tests as you go will make you less prone to this problem, we can talk more about that if you like.

> In retrospect it would be better if the parler_po_export command wrote a single file to stdout by default, which would make it easier to test properly (and then we could add a command that wraps that for creating a directory structure as with makemessages). 

I mostly agree. It is usually the best to isolate business logic (that require a lot of tests) from code that handles how data gets into and out of the system. That way your business logic API can be designed with ease of testing as *the primary driver* of the design. This usually means that business logic that generates text data would return that as a string, and some other part of the system would handle pushing that string to standard out or whatever.

> But, alas. 

There is no alas, we will revisit this code as we work on this part of the functionality, and each time is a chance to make things a little bit better, and applying more of what we've learned about our domain, etc.

> As a result, the tests for the commands themselves are pretty high level integration tests. But on the bright side, they'll help us catch things we probably wouldn't otherwise.

There is nothing wrong with having integration tests, as long as what they test is the integration points, not all the nitty gritty of the business logic. But we can leave this for future refactoring. Since there are tests, we are in a position to fix this down the line. Without tests, we would probably not be.

parler_po/argparse_path.py
```diff
@@ -0,0 +1,57 @@
...
+        self._file_type = file_type
+        self._mode = mode
+
+    def __call__(self, *args):
+        mode_flags = set(DIR_MODE_FLAGS[k] for k in self._mode)
+        return _path_type_fn(
+            *args, file_type=self._file_type, mode_flags=mode_flags
+        )
+
+def _path_type_fn(path, file_type=None, mode_flags=set()):
```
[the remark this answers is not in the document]
dylanmccall: This function is modelled after argparse.FileType. It's essentially that, but for both directories and files. I think for the purpose of code like that it's clearer to build it with a familiar interface, and ideally move it to an external module. (It isn't right now, but that could be fixed).

There was an error earlier where I misnamed a couple variables, making it look like this only did half of what its name suggested, and that has been fixed.

config/settings/base.py
```diff
@@ -42,6 +42,7 @@
     'organizations.apps.OrganizationsConfig',
     'search.apps.SearchConfig',
     'users.apps.UsersConfig',
+    'parler_po.apps.ParlerPOConfig'
```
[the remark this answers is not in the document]
dylanmccall: Oh yes, that! Sure, we can rename this. I think it's good to have names that describe our core technical assumptions, but using Parler definitely isn't one of those. (The code to get translations for some record is pretty loosely attached and we could easily change it). That this deals with PO files is a little more debatable.

How's `content_translations_sync`?

parler_po/exceptions.py
dylanmccall: Fixed. Thanks :)

parler_po/exceptions.py
dylanmccall: Yeah, I don't know if we really reached a conclusion on that? I'd pointed out that folks generally mix the two in Python depending on the semantics, and then Github probably ate the comment thread because it was attached to a change somewhere :) No problem doing `s/\"/\'/g`.

parler_po/tests/helpers.py
dylanmccall: I tidied up the use of builder objects here in my most recent commit. Should help a little with stuff like this.

parler_po/tests/helpers.py
dylanmccall: I tidied up the use of builder objects here in my most recent commit. Should help a little with stuff like this.

parler_po/tests/test_field_ids.py
dylanmccall: Those are from our own module :) They're to be considered private by other code, but tests are special.

dylanmccall:
Okay, you should be much happier with this branch now, Rasmus :)

rasmus-storjohann-PG:
Just doing a QA sweep of this feature now:

The screen output from the BC211 data import doesn't look right, `<built-in function id>` shouldn't be there

```
INFO parser Parsed taxonomy term: <built-in function id> YZ-1700
INFO parser Parsed taxonomy term: <built-in function id> {what:'ADVOCACY',why:'compensation',why:'injury',who:'Employers'}
```

The logging of parsing of taxonomy terms is also verbose and slows down importation, about 90% of log output is related to taxonomy, can it be changed from INFO to DEBUG?

rasmus-storjohann-PG:
Please add this to the main README file:

```
This adds some management commands under the parler_po application.

List available translatable models:

./manage.py content_translation_list_models

Create a POT file for a particular translatable model:

./manage.py content_translation_export organizations.organization > organization.pot

Create a PO file for a particular translatable model in French:

./manage.py content_translation_export organizations.organization --language=fr > fr/organization.po
The resulting files can be opened in an application such as GTranslator or Poedit.

After you have updated your translations, you can import them again:

./manage.py content_translation_import fr/translation.po
```

rasmus-storjohann-PG:
Doing an export of locations and organizations, I see a lot of empty msgid and msgstr fields, a typical example, is this not a bug? 

```
#: organizations.organization@description@9490996
msgid ""
"Offers 12-month, faith-based residential programs for men and women aged 19 "
"and over who are struggling with addiction or any other life-controlling "
"problem. Also offers a vehicle donor program. Office and intake hours are 9 "
"am to 5 pm Monday to Friday. Nonprofit society."
msgstr ""
```

dylanmccall:
Okay, README is updated, and the unrelated taxonomies thing was done in e66fbef.

rasmus-storjohann-PG:
All right, @dylanmccall, it's merged! Thanks for doing all that work.

rasmus-storjohann-PG:
@bensinjin Can you please review the discussion that Dylan and I had on this PR. There's quite a bit there that will apply generally for future PRs too.

dylanmccall:
Great! Thanks for your patience, and for merging :)

bensinjin:
@rasmus-storjohann Will do.

PR #146, opened 2017-12-06
Issue #138 Add a search application and viewset

### LINKED ISSUE #138: search: create API end point that returns all data
AC: GET request to the search end point returns all organizations from the database

Possible dupe of #93, clean up #93 if so

dylanmccall:
The search application provides a "/search" API that returns data according to the HSDA Search specification. For now it only returns a list of organizations, without filtering.

Closes #138.


search/tests/test_api.py
```diff
@@ -0,0 +1,51 @@
...
+
+class ServiceProvidersApiTests(rest_test.APITestCase):
+    def test_lists_types(self):
+        url = '/v1/search/'
+        response = self.client.get(url)
+        response_json = response.json()
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertIn('organizations', response_json)
+        self.assertIn('locations', response_json)
+        self.assertIn('services', response_json)
```
rasmus-storjohann-PG: If you break out the three first lines into a helper function, then you can turn this one test into four two line tests, which will be nice

search/tests/test_api.py
```diff
@@ -0,0 +1,51 @@
...
+    def test_can_get_organizations(self):
+        name = 'the_name'
+        organization = OrganizationBuilder().with_name(name).build()
+        organization.save()
+        url = '/v1/search/'
+        response = self.client.get(url)
+        response_json = response.json()
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response_json['organizations']), 1)
+        self.assertEqual(response_json['organizations'][0]['name'], name)
```
rasmus-storjohann-PG: To really prove that you can get organization*s*, you can create two organizations and make sure you get two back

dylanmccall:
Okay, those are taken care of now :)

mysite/urls.py
```diff
@@ -1,15 +1,16 @@
...
 
 def build_router():
     router = routers.DefaultRouter()
-    router.register(r'^service-providers', service_providers_viewsets.ServiceProviderViewSet)
-    router.register(r'^organizations', organizations_viewsets.OrganizationViewSet)
-    router.register(r'^search', search_viewsets.SearchViewSet)
+    router.register(r'^organizations', OrganizationViewSet)
+    router.register(r'^organizations/(?P<organization_id>[0-9a-zA-Z_]+)/locations', LocationViewSetUnderOrganizations, 'location')
+    router.register(r'^locations', LocationViewSet, 'location')
+    router.register(r'^search', SearchViewSet)
```
rasmus-storjohann-PG: This commit is funky, I don't think the rebase was done quite right. I would check out current master and cherry-pick your two commits to fix this, but there may be better ways to fix this.

Issue #106, opened 2017-09-29
Establish pattern for user account, login, session


rasmus-storjohann:
AC: Can create use accounts, set password, log in, log out. This works in local, test and production environments running locally, and in production on Heroku. 

Dead code from [cookiecutter-django](http://cookiecutter-django.readthedocs.io/) (merged in #151) has been cleaned up.

| Environment | Status |
|---|---|
| local on localhost | Can sign up and log in, confirmation URL sent to console |
| test on localhost | Can sign up, confirmation sent by email, which does not work |
| production on localhost | Cannot run server, SSL error |
| test on Travis | |
| production on Heroku | |

rasmus-storjohann-PG:
See 399d4e72a5316ea5ead78e279a02c9d6e030ad0c

PR #151, opened 2017-12-19
Feature/issue 113 establish deployment process

rasmus-storjohann-PG:


rasmus-storjohann-PG:
OK, @dylanmccall I think this one is ready to go now. There is still quite a bit of dead code related to accounts in here, but we can remove that as we go or revive it.

dylanmccall:
Sure. Yeah, probably feels like a bit of a slog at this point. Looks fine, so I'll merge it. Thanks for taking care of this! I added that cleanup stuff as an acceptance criteria in #106.

PR #152, opened 2017-12-20
Feature/issue 128 Service: Build data model and read-only API

### LINKED ISSUE #128: service: Build data model and read-only API
Precondition: organization and location data models are in place.

Acceptance Criteria:

* When importing data from BC211 XML input, all fields are included as per the tables below, the service_at_location table being populated to indicate which location each service is available at. 
* A GET request to /services/ returns a list of all services with all the fields populated
* A GET request to /services/{id} returns the services with the given id
* A GET request to /organizations/{id}/services/ returns a list of all services under the given organization
* A GET request to /organizations/{id}/services/{id} returns the services with the given organization and service ids

One small unknown: Given the BC211 XML format with the service being child nodes of locations, either each service is available at one location only, or else the services are duplicated under each location where they are available. This should not be too hard to figure out once we have a parser.

service: 

Name | BC211 XPath | Translatable
-- | -- | --
id   | Source/Agency/Site/SiteService/Key | No
organization_id   |  Source/Agency/Key | No
name |   Source/Agency/Site/SiteService/Name | Yes
description |   Source/Agency/Site/SiteService/Description  | Yes

service_at_location:

| Name |
|----|
| service_id |
| location_id |

dylanmccall:
Closes #128.


config/urls.py
```diff
@@ -11,10 +11,10 @@
 
 def build_router():
     router = routers.DefaultRouter()
-    router.register(r'^organizations', OrganizationViewSet)
-    router.register(r'^organizations/(?P<organization_id>[0-9a-zA-Z_]+)/locations', LocationViewSetUnderOrganizations, 'location')
-    router.register(r'^locations', LocationViewSet, 'location')
-    router.register(r'^search', SearchViewSet)
+    router.register(r'organizations', OrganizationViewSet, base_name='organization')
```
rasmus-storjohann-PG: Fix commit message to include issue number.

I'm curious why it's better to not have the ^

locations/models.py
```diff
@@ -27,6 +29,13 @@ def validate_latitude_and_longitude(self):
         if latitude_is_null != longitude_is_null:
             raise_mismatch_exception(latitude_is_null, longitude_is_null)
 
+class ServiceLocation(ValidateOnSaveMixin, TranslatableModel):
```
rasmus-storjohann-PG: The story calls this table `ServiceAtLocation` or `service_at_location`, and there is no `description` field.

locations/tests/test_models.py
```diff
@@ -135,3 +137,34 @@ def set_description_in_language(self, location, language, text):
...
+        self.service = ServiceBuilder(self.organization).build()
+        self.service.save()
+
+        self.location = LocationBuilder(self.organization).build()
+        self.location.save()
+
+    def test_has_service_field(self):
+        service_location = ServiceLocationBuilder(self.service, self.location).build()
+        service_location_from_db = validate_save_and_reload(service_location)
+        self.assertEqual(service_location_from_db.service, self.service)
```
rasmus-storjohann-PG: I'm curious what is actually compared here, since you're comparing two objects.

locations/tests/helpers.py
```diff
@@ -38,3 +38,20 @@ def build(self):
         result.longitude = self.longitude
         result.description = self.description
         return result
+
+class ServiceLocationBuilder:
+    def __init__(self, service, location):
+        self.service = service
+        self.location = location
+        self.description = 'default description'
```
rasmus-storjohann-PG: The value of the builder pattern is to not have to specify default values for a bunch of properties. Here, you have to specify values for the only two properties that matter, so there is no point in using a builder. I want us to follow established patterns in the code base, but only when they make sense to use.

services/models.py
```diff
@@ -0,0 +1,26 @@
...
+from organizations.models import Organization
+from parler.models import TranslatableModel, TranslatedFields
+
+class Service(ValidateOnSaveMixin, TranslatableModel):
+    id = RequiredCharField(primary_key=True, max_length=200, validators=[validators.validate_slug])
+    organization = models.ForeignKey(Organization, on_delete=models.CASCADE, related_name='services')
+    translations = TranslatedFields(
+        name=models.CharField(max_length=200),
+        description=models.TextField(blank=True, null=True),
+        location_description=models.TextField(blank=True, null=True)
```
rasmus-storjohann-PG: There is no field `location_description` in the story

services/serializers.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework import serializers
+from services import models
+from locations.serializers import LocationSerializer
+
+class ServiceSerializer(serializers.HyperlinkedModelSerializer):
```
rasmus-storjohann-PG: If you can figure out how to get hyperlinks in the resulting DTOs, that would be cool. I tried for a bit a while ago

services/tests/test_models.py
```diff
@@ -0,0 +1,120 @@
...
+        service_from_db = validate_save_and_reload(service)
+        self.assertEqual(service_from_db.description, null_description)
+
+    @unittest.expectedFailure
+    def test_empty_description_is_saved_as_null(self):
+        empty_description = ''
+        null_description = None
+        service = ServiceBuilder(self.organization).with_description(empty_description).build()
+        service_from_db = validate_save_and_reload(service)
+        self.assertEqual(service_from_db.description, null_description)
```
rasmus-storjohann-PG: Did you try to make this work?

services/tests/test_models.py
```diff
@@ -0,0 +1,120 @@
...
+        service = ServiceBuilder(self.organization).build()
+        service_from_db = validate_save_and_reload(service)
+
+        location = LocationBuilder(self.organization).build()
+        location_from_db = validate_save_and_reload(location)
+
+        service_location = ServiceLocationBuilder(service, location).build()
+        service_location_from_db = validate_save_and_reload(service_location)
+
+        self.assertEqual(service.location, location)
```
rasmus-storjohann-PG: Better names of your local variables will make this test easier to understand. I think you also need to have >1 locations in the db to prove that the one you get back is the first.

services/models.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+
+    def __str__(self):
+        return self.name
+
+    @cached_property
+    def location(self):
+        return self.locations.first()
```
rasmus-storjohann-PG: Why is this needed? It's a little problematic because the API creates the strong impression that there is one location for a given service, when actually there are many. I would prefer a `service.locations` property instead.

config/urls.py
dylanmccall: Oops, yeah, that one just doesn't have an issue associated. I'll make one and assign it to that commit. (And may as well merge it from another branch while I'm at it). The reason there is router.register takes a prefix, not a full URL pattern. It probably adds its own start of string regex character somewhere.

services/serializers.py
dylanmccall: That's what that earlier misplaced commit fixes :)

services/tests/test_models.py
dylanmccall: I filed an issue report: #153.

services/models.py
dylanmccall: Indeed. Okay, moved to the API. It's needed there becuase HSDA wants a single location for a service, but we don't need it in the model layer.

config/urls.py
rasmus-storjohann-PG: Don't worry about that, it can be part of this issue, I'd rather not have commits related to different issues in the same PR

services/serializers.py
rasmus-storjohann-PG: Because I'd love to have urls in DTOs, REST all the way!!!

rasmus-storjohann-PG:
As to that location property on the service DTO, did you see you got a response on your issue against HSDS? The field is deprecated, so you can leave it if it makes sense

services/tests/test_models.py
rasmus-storjohann-PG: Thanks. To fix this I need a way to iterate over translated fields, which seems to be totally doable. In addition, we'll also need to give a bit of a think to whether NOT NULL is a meaningful constraint on translated fields, and what the semantics of that would be...

config/urls.py
dylanmccall: I don't like mixing issues either :b Cherry-picked into a new branch for issue #156, with an issue number in the commit message.

dylanmccall:
I missed a spot, so this has a few other changes that need review.

rasmus-storjohann-PG:
Merge conflicts on this branch now :-(

Issue #147, opened 2017-12-06
search: populate many to many relation between taxonomy terms and services


dylanmccall:
**Table: service_taxonomy**

| Name | Description |
|---|---|
| id | PK, may not be needed |
| service_id | FK |
| taxonomy_id | FK |

See https://openreferral.readthedocs.io/en/latest/hsds/reference/#service_taxonomy

The id field will not be needed for the API (you will never refer directly to a linking relation in the API), so my preference is to use composite primary keys on linking relations like this and not have the id field. To be discussed, I'm sure.

Populate the tables from BC211 taxonomy. The simplest way may be to do this as part of the parsing of the BC211 XML data: As each service is parsed, take the following

```
<Taxonomy>
    <Code>TJ-4400.8100</Code>
</Taxonomy>
<Taxonomy>
    <Code>{what:'RECREATION',why:'disability',who:'Children'}</Code>
</Taxonomy>
```

ignore entries of the type 'TJ-4400.8100' and generate the following records (if they don't already exist) in the `taxonomy` table. Convert terms to lower case and replace space with underscore to be nice:

| id | taxonomy_id | name |
|---|---|---|
| id1 | 'what' | 'recreation' |
| id2 | 'why' | 'disability' |
| id3 | 'who' | 'children' |

and also generate records in the `service_taxonomy` table connecting these taxonomy terms to the service being parsed. Alternatively, the taxonomy tables can be imported from some other reliable source and the linking relations made at the time of parsing the services data.

PR #164, opened 2018-01-11
Feature/issue 147

bensinjin:
This closes issue #147. When I got this issue the taxonomy term table and import for terms were already in place. The work done in this branch implements the many to many relationship from services to taxonomy terms, tests and tweaks the import to populate this association on import.


bc211/importer.py
```diff
@@ -93,23 +93,23 @@ def save_services(services, counters):
...
-            LOGGER.debug('Imported taxonomy term: %s %s', taxonomy_term.taxonomy_id, taxonomy_term.name)
-
-def save_taxonomy_term(record):
-    active_record, created = TaxonomyTerm.objects.get_or_create(
-        taxonomy_id=record.taxonomy_id,
-        name=record.name
-    )
-    return created
+            LOGGER.info('Imported taxonomy term: %s %s', taxonomy_term.taxonomy_id, taxonomy_term.name)
+        service_active_record.taxonomy_terms.add(taxonomy_term_active_record)
```
dylanmccall: It would be lovely to see a test for this new functionality. A good spot for that is in [bc211/tests/test_importer.py](https://github.com/pg-irc/pathways-backend/blob/feature/issue_147/bc211/tests/test_importer.py). Perhaps you could add a ServiceImportTests class there like the existing OrganizationImportTests (sorry, I should have done that when I added services!), and then you can just check that the first service has the correct set of taxonomy terms.

bensinjin: Yeah I did think of that. I'll work on some tests for this, cheers!

services/tests/test_models.py
```diff
@@ -118,3 +119,23 @@ def set_description_in_language(self, service, language, text):
...
+        service = validate_save_and_reload(ServiceBuilder(self.organization).build())
+        taxonomy_term_one = validate_save_and_reload(TaxonomyTermBuilder().with_name('one').build())
+        taxonomy_term_two = validate_save_and_reload(TaxonomyTermBuilder().with_name('two').build())
+
+        service.taxonomyTerms.add(taxonomy_term_one)
+        service.taxonomyTerms.add(taxonomy_term_two)
+        service.taxonomyTerms.remove(taxonomy_term_one)
+
+        self.assertEqual(service.taxonomyTerms.count(), 1)
+        self.assertEqual(service.taxonomyTerms.first(), taxonomy_term_two)
```
rasmus-storjohann-PG: The two asserts here together are really saying the collection should contain just this one specific element. Tests are improved by using asserts that closely match what you're after, which also gets us closer to the ideal of one assert per test. So if you were to keep this test, then use https://docs.python.org/3.2/library/unittest.html#unittest.TestCase.assertCountEqual

bc211/importer.py
```diff
@@ -93,23 +93,22 @@ def save_services(services, counters):
...
-def save_taxonomy_term(record):
-    active_record, created = TaxonomyTerm.objects.get_or_create(
-        taxonomy_id=record.taxonomy_id,
-        name=record.name
-    )
-    return created
+        taxonomy_term_active_record, created = TaxonomyTerm.objects.get_or_create(
+            taxonomy_id=taxonomy_term.taxonomy_id,
+            name=taxonomy_term.name
+        )
```
rasmus-storjohann-PG: I think it will be cleaner to break the `TaxonomyTerm.objects.get_or_create()` and `counters.count_taxonomy_term()` calls out into a separate function

bc211/importer.py
bensinjin: A test for this is now in place.

bc211/importer.py
bensinjin: Makes sense. See 8bf9bc84d0a9fb4857e30fba1fe19040b48872fb.

services/tests/test_models.py
```diff
@@ -118,3 +119,23 @@ def set_description_in_language(self, service, language, text):
     def assert_description_in_language_equals(self, service, language, expected_text):
         service.set_current_language(language)
         self.assertEqual(service.description, expected_text)
+
+    def test_can_add_taxonomy_term(self):
+        service = validate_save_and_reload(ServiceBuilder(self.organization).build())
```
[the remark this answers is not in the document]
bensinjin: I've added a method to our builder. See 7101ac0c68cc34f585b5317894d768e21f2b07b9.

services/tests/test_models.py
```diff
@@ -118,3 +119,23 @@ def set_description_in_language(self, service, language, text):
...
+        self.assertEqual(service.taxonomyTerms.first(), taxonomy_term)
+
+    def test_can_remove_taxonomy_term(self):
+        service = validate_save_and_reload(ServiceBuilder(self.organization).build())
+        taxonomy_term_one = validate_save_and_reload(TaxonomyTermBuilder().with_name('one').build())
+        taxonomy_term_two = validate_save_and_reload(TaxonomyTermBuilder().with_name('two').build())
+
+        service.taxonomyTerms.add(taxonomy_term_one)
+        service.taxonomyTerms.add(taxonomy_term_two)
+        service.taxonomyTerms.remove(taxonomy_term_one)
```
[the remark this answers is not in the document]
bensinjin: I've removed this test as I've deemed it more of a test of how Django should work. See 7101ac0c68cc34f585b5317894d768e21f2b07b9.

services/tests/test_models.py
bensinjin: See previous comment. This test has been removed but thank you for the feedback. I've reviewed the documentation and will be more carefully selecting my future asserts.

services/tests/helpers.py
```diff
@@ -19,10 +20,17 @@ def with_description(self, description):
...
+
     def build(self):
         result = models.Service()
         result.id = self.id
         result.name = self.name
         result.organization = self.organization
         result.description = self.description
+        if self.taxonomy_terms:
+            for taxonomy_term in self.taxonomy_terms:
+                result.taxonomy_terms.add(taxonomy_term)
```
rasmus-storjohann-PG: Not too worried that someone will be calling `with_taxonomy_terms(None)`, so you can remove the `if` clause

services/tests/test_models.py
```diff
@@ -6,6 +6,7 @@
 from services.tests.helpers import ServiceBuilder
 from organizations.tests.helpers import OrganizationBuilder
 from locations.tests.helpers import LocationBuilder, ServiceLocationBuilder
+from taxonomies.tests.helpers import TaxonomyTermBuilder
```
rasmus-storjohann-PG: This can be removed

services/tests/helpers.py
bensinjin: See 83e79c6492ed20405410699263217ceb8dc4692d.

services/tests/test_models.py
bensinjin: See 83e79c6492ed20405410699263217ceb8dc4692d.

PR #172, opened 2018-01-18
Feature/issue 140 taxonomic search

rasmus-storjohann-PG:


services/tests/test_api.py
```diff
@@ -2,29 +2,30 @@
...
 from organizations.tests.helpers import OrganizationBuilder
+from common.testhelpers.autofixtures import *
 
 class ServicesApiTests(rest_test.APITestCase):
     def setUp(self):
-        self.organization_id = 'the_organization_id'
+        self.organization_id = a_string()
         self.organization = OrganizationBuilder().with_id(self.organization_id).build()
         self.organization.save()
         self.data = {
```
bensinjin: These data values should be being used in the other tests in this file instead of re calling a_string().

bensinjin: These data values should be being used in the other tests in this file instead of re calling a_string().

bensinjin: These data values should be being used in the other tests in this file instead of re calling a_string().

bensinjin: These values should be used throughout ServicesApiTests as apposed to re calling a_string().

rasmus-storjohann-PG: self.data is used for post and put calls only.

PR #174, opened 2018-01-23
Feature/issue 165 full text search

rasmus-storjohann-PG:


services/tests/test_search.py
```diff
@@ -64,15 +60,36 @@ def test_search_with_wrong_taxonomy_id_returns_404(self):
...
+        self.assertEqual(response.json()[0]['name'], the_name)
+
+    def test_full_text_search_with_wrong_search_term_returns_404(self):
+        the_wrong_name = a_string()
+        service = ServiceBuilder(self.organization).create()
+
+        url = '/v1/services/?queries={0}'.format(the_wrong_name)
+        response = self.client.get(url)
+
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
```
bensinjin: See my notes on the issue. 404 doesn't seem right here.

services/models.py
```diff
@@ -38,3 +39,10 @@ def add_taxonomy_filter_if_given(queryset, search_parameters):
             queryset = queryset.filter(taxonomy_terms__taxonomy_id=identifier,
                                        taxonomy_terms__name=term)
         return queryset
+
+    @staticmethod
+    def add_full_text_search_filter_if_given(queryset, search_parameters):
+        if search_parameters.full_text_search_terms:
```
bensinjin: It feels like this conditional (and probably the ones in add_taxonomy_filter_if_given) shouldn't be the responsibilities of these methods. You could just call this add_full_text_search_filter and do the conditional logic in another method/function or just prior to actually adding the filter.

services/models.py
```diff
@@ -44,8 +44,15 @@ def add_taxonomy_filter_if_given(queryset, search_parameters):
...
-                                       Q(translations__description__contains=search_term))
+            qs = None
+            search_terms = search_parameters.full_text_search_terms
+            for search_term in search_terms:
+                if qs:
+                    qs = qs | Q(translations__name__contains=search_term)
+                    qs = qs | Q(translations__description__contains=search_term)
+                else:
+                    qs = Q(translations__name__contains=search_term)
+                    qs = qs | Q(translations__description__contains=search_term)
```
bensinjin: This seems quite redundant, probably want a function (maybe even just an inner function if you don't need it elsewhere) to build your .. query string (assuming that's what qs is)? Probably be nice to rename qs to something clearer. Can you also explain what the | is doing here?

services/details.py
```diff
@@ -26,4 +26,9 @@ def parse_full_text_search_terms(query_parameters):
     query_arguments = query_parameters.get('queries', None)
     if not query_arguments:
         return None
-    return query_arguments.split(' ')
+    result = query_arguments.split(' ')
+    cleaned_result = []
+    for x in result:
+        if x:
+            cleaned_result.append(x)
+    return cleaned_result
```
bensinjin: I see you cleaned this up later.

bensinjin:
This is looking pretty good. I had some comments I've added to the ticket here: https://github.com/pg-irc/pathways-backend/issues/165

services/tests/test_search.py
rasmus-storjohann-PG: Fixed

services/models.py
bensinjin: Makes sense.

services/models.py
bensinjin: Nice solution

PR #178, opened 2018-01-27
Feature/issue 134

### LINKED ISSUE #134: physical_address: Build data model and read-only API
The design of these tables are based on http://www.databaseanswers.org/data_models/customers_and_addresses/index.htm since the schema in HSDS involves duplication of tables and inflexible key relationships.

Pattern to follow: the organizations and locations end points. 
Create `pathways-backend/addresses` to hold the code and `addresses` and `address_types` database tables. The `location_addresses` table should be defined in `pathways-backend/locations`.

Acceptance criteria:

* The `address_types` table is prepopulated with the values "physical_address" and "postal_address".
* When importing data from BC211 XML input, all fields are included as per the table below, the attention and region fields will be null. The id field is generated by the DB as per the django default behaviour, as it is not present in the BC211 data. The `location_addresses` table is populated to relate each location to their respective physical and postal addresses.
* When the identical address appears more than once in the input, only one such record is created in the `addresses` table.
* For addresses that contain more than one line of address information (`LineX` elements exist for X>1), the line elements should be concatenated with newline separation and the result stored in the  `address` field.
* Any missing element should be stored as NULL in the database, not the empty string. Saving an address with an empty string for any field should convert the empty string to NULL.
* The database table definition should contain NOT NULL constraints reflecting the information below.
* The database table definition contains a UNIQUE TOGETHER constraint on `location_addresses.location_id` and `location_addresses.address_type_id` to ensure that each location only has one of each type of address.
* Do either (based on what is your preference as a consumer of this API):
** A GET request to  `/locations/{id}/physical_address` returns the physical address of the location with the given id, or 404 if there is no such address, OR
** A GET request to  `/locations/{id}/postal_address` returns the postal address of the location with the given id, or 404 if there is no such address.
** The response from a GET request to `/locations/{id}` contains the primary key to the entities for the postal or physical address. If either address is not present, it is represented by `null` as per https://stackoverflow.com/questions/21120999/representing-null-in-json.
* Or
** Change the response from the `/locations/` endpoint to include the addresses of both types embedded in the JSON response for the location. If either address is not present, it is represented by `null`.
* Validation on the country field should ensure that it contains a two-character all upper case string, see  ISO 3361-1 standard.

Fields included (none of these fields are translatable):

## Table: addresses

| Name | BC211 XPath |Can be null?|
|--- | --- | --- |
id| | No |
attention | | Yes |
address|Source/Agency/Site/PhysicalAddress/Line1 through Line4 | No |
city|Source/Agency/Site/PhysicalAddress/City | No |
state_province|Source/Agency/Site/PhysicalAddress/State | Yes |
postal_code|Source/Agency/Site/PhysicalAddress/ZipCode | Yes |
country|Source/Agency/Site/PhysicalAddress/Country | No |

## Table: address_types

|Name|Can be null?|
|--- | --- |
| id | No |

## Table: location_addresses

|Name| Can be null?|Unique together|
|--- | --- | --- |
|id| No | |
|address_id| No | |
|location_id| No | Yes |
|address_type_id| No | Yes |

bensinjin:
You probably only need to review commits after "#134 Cleanup address parsing tests and use existing dummy data … (fb1863ea068cae4d22b1f5888de4a86f2434ce59)".  Generally speaking commits prior were fixes a pertaining to your previous review of older commits, naturally you can go back as far as you'd like.

I have opted to embed addresses for locations as this feels better to me.  This closes #134.


bc211/tests/test_importer.py
```diff
@@ -75,6 +76,37 @@ def test_service_has_correct_taxonomy_terms(self):
...
+        file = open(ONE_AGENCY_FIXTURE, 'r')
+        records = read_records_from_file(file)
+        save_records_to_database(records)
+        self.addresses = Address.objects.all()
+
+    def test_can_import_address(self):
+        self.assertIsInstance(self.addresses.first(), Address)
+
+    def test_does_not_import_duplicates(self):
+        self.assertEqual(len(self.addresses), 1)
```
rasmus-storjohann-PG: The only way I can make sense of this test is to look at `ONE_AGENCY_FIXTURE` in detail. I prefer tests that I can read in isolation.

bc211/tests/test_importer.py
```diff
@@ -75,6 +76,37 @@ def test_service_has_correct_taxonomy_terms(self):
...
+
+class AddressTypeTests(TestCase):
```
rasmus-storjohann-PG: Rather than populating the address type table from the BC211 data, I'd like to see something in the `model.py` file to create the two records it needs and then have the importer throw if any other kinds of addresses show up. This means that this test really just needs to check that those two records are the only ones that exist.

bc211/tests/test_parser.py
```diff
@@ -37,6 +38,21 @@
...
+            <MailingAddress>
+                <Line2>Line2</Line2>
+                <Line1>Line1</Line1>
+                <City>City</City>
+                <Country>Country</Country>
+                <State>State</State>
+                <ZipCode>Code</ZipCode>
+            </MailingAddress>
+            <PhysicalAddress>
+                <Line1 />
```
rasmus-storjohann-PG: So this address is invalid, right? I'd expect to see just valid addresses here so that someone reading the tests below doesn't need to remember/look up/be confused by the fact that one of these addresses is invalid. Tests that check for the behaviour on invalid tests should contain the invalid data closer to the test, I think.

bc211/tests/test_parser.py
```diff
@@ -51,6 +67,21 @@
...
```
rasmus-storjohann-PG: Same comment as above regarding invalid data in the fixture.

bc211/tests/test_parser.py
```diff
@@ -194,3 +225,19 @@ def test_sets_the_site_id(self):
...
+class AddressParserTests(unittest.TestCase):
+    def setUp(self):
+        root = etree.fromstring(MINIMAL_211_DATA_SET)
+        site = root.find('Agency/Site')
+        self.postal_address = site.find('MailingAddress')
+        self.physical_address = site.find('PhysicalAddress')
+
+    def test_only_parses_valid_addresses(self):
+        parsed_address = parser.parse_address(self.physical_address,
+                                              '12345', 'physical_address')
```
rasmus-storjohann-PG: So what is `12345` doing here? I think these tests will probably be easier to follow if the XML is right in the test.

bc211/tests/test_parser.py
```diff
@@ -194,3 +225,19 @@ def test_sets_the_site_id(self):
...
+        self.assertEqual(parsed_address, None)
+
+    def test_parsed_address_lines_correctly_formatted(self):
+        address_lines = parser.parse_address_lines(self.postal_address)
+        self.assertEqual(address_lines, 'Line1\nLine2')
```
rasmus-storjohann-PG: The function that is doing all the assembly of lines does way more than this test covers, including what if line1 is empty, lines all the way up to 9 (you can change that and assume that we will never see lines above 4, as long as there is an error if we ever do see a line5 or higher so we notice.

config/settings/base.py
```diff
@@ -44,7 +44,8 @@
     'taxonomies.apps.TaxonomiesConfig',
     'search.apps.SearchConfig',
     'users.apps.UsersConfig',
-    'content_translation_tools.apps.ContentTranslationToolsConfig'
+    'content_translation_tools.apps.ContentTranslationToolsConfig',
+    'addresses.apps.AddressesConfig'
 ]
```
rasmus-storjohann-PG: See how the diff here is 2 lines when you really only changed one? Python fixes that problem by allowing you to have `,` after the last element in a list, so next time, the diff will be just one line. I haven't been using that feature much, but I think we should.

locations/models.py
```diff
@@ -43,6 +47,27 @@ def __str__(self):
...
+
+class LocationAddress(ValidateOnSaveMixin, models.Model):
+    address = models.ForeignKey(Address,
+                                related_name='address_locations',
+                                on_delete=models.CASCADE)
+    location = models.ForeignKey(Location,
+                                 related_name='location_addresses',
+                                 on_delete=models.CASCADE)
+    address_type = models.ForeignKey(AddressType,
+                                     on_delete=models.CASCADE)
```
rasmus-storjohann-PG: Not sure if both of these `related_name` fields should be there... Also we need to think kinda hard about the `on_delete` values across all our models.

locations/serializers.py
```diff
@@ -1,7 +1,18 @@
...
+        fields = ('address_type', 'address')
+
+
+class LocationSerializer(serializers.ModelSerializer):
+    location_addresses = LocationAddressSerializer(many=True)
     class Meta:
         model = models.Location
-        fields = ('id', 'name', 'organization_id', 'latitude', 'longitude', 'description')
+        fields = ('id', 'name', 'organization_id', 'latitude',
+                  'longitude', 'location_addresses', 'description')
```
rasmus-storjohann-PG: So the location JSON here will have a `location_addresses` object with two properties `physical_address` and `postal_address`? Would it be possible to have the `physical_address` and `postal_address` objects directly in the location without the `location_addresses` object?

locations/tests/helpers.py
```diff
@@ -52,3 +52,19 @@ def build(self):
         result.service = self.service
         result.location = self.location
         return result
+
+class LocationAddressBuilder:
+    def __init__(self, address, location, address_type):
+        self.address = address
+        self.location = location
+        self.address_type = address_type
```
rasmus-storjohann-PG: When all the properties need to be specified in arguments to the builder constructor, it means that you don't need the builder, you can just construct the DTO directly. Builders are only helpful when they define default values for fields that most tests don't care about.

locations/tests/test_api.py
```diff
@@ -86,3 +90,23 @@ def test_cannot_delete_on_organization(self):
...
+        AddressTypeBuilder().with_id(address_type_id).create()
+        LocationAddressBuilder(
+            location=Location.objects.first(),
+            address=Address.objects.first(),
+            address_type=AddressType.objects.first()
+        ).create()
+        url = '/v1/locations/'
+        response = self.client.get(url)
+        location_addresses = response.json()[0]['location_addresses']
+        self.assertEqual(location_addresses[0]['address_type'], address_type_id)
```
rasmus-storjohann-PG: Let's discuss.

addresses/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
```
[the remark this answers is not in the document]
bensinjin: Cool. I've included the addresses in the admin site.

addresses/serializers.py
```diff
@@ -0,0 +1,15 @@
+from rest_framework import serializers
+from addresses import models
+
+
+class AddressSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.Address
+        fields = ('id', 'address', 'city', 'country',
+                  'attention', 'state_province', 'postal_code')
```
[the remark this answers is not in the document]
bensinjin: I've rearranged these fields (on serializer and model) to match the story. Originally my reasoning was to put the required fields first but I'm not married to that approach.

addresses/tests/helpers.py
```diff
@@ -0,0 +1,65 @@
+from addresses import models
+from common.testhelpers.random_test_values import a_string
+
+class AddressBuilder:
+    def __init__(self):
+        self.address = a_string()
+        self.city = a_string()
+        self.country = 'CA'
```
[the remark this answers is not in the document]
bensinjin: Please take a look at https://github.com/pg-irc/pathways-backend/commit/724c15b3f640d2bc0c39ba8a6a22f82db33422a3. This allows us further flexibility with a_string as well as adding a few tests for the added functionality.

addresses/tests/helpers.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def with_postal_code(self, postal_code):
+        self.postal_code = postal_code
+        return self
+
+    def create(self):
+        self.build().save()
+
+
+class AddressTypeBuilder:
```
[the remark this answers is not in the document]
bensinjin: Agreed, fixed.

addresses/tests/test_models.py
```diff
@@ -0,0 +1,61 @@
...
+        self.assertEqual(address_from_db.city, address.city)
+
+    def test_has_country_field(self):
+        country_field_value = 'NZ'
+        address = AddressBuilder().with_country(country_field_value).build()
+        address_from_db = validate_save_and_reload(address)
+        self.assertEqual(address_from_db.country, address.country)
+
+    def test_only_allow_two_character_country_fields(self):
+        country_field_value = a_string()
```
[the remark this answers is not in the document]
bensinjin: Agreed and fixed. Changes to the a_string make this nicer.

addresses/tests/test_models.py
```diff
@@ -0,0 +1,61 @@
...
+        state_province_field_value = a_string()
+        address = AddressBuilder().with_state_province(state_province_field_value).build()
+        address_from_db = validate_save_and_reload(address)
+        self.assertEqual(address_from_db.state_province, address.state_province)
+
+    def test_has_postal_code_field(self):
+        postal_code_field_value = a_string()
+        address = AddressBuilder().with_postal_code(postal_code_field_value).build()
+        address_from_db = validate_save_and_reload(address)
+        self.assertEqual(address_from_db.postal_code, address.postal_code)
```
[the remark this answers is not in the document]
bensinjin: Fixed.

bc211/importer.py
```diff
@@ -24,12 +26,15 @@ def count_location(self):
...
-        self.service_at_location_count += 1
-
     def count_taxonomy_term(self):
         self.taxonomy_term_count += 1
 
+    def count_address(self):
+        self.address_count += 1
+
+    def count_address_type(self):
+        self.address_type_count += 1
```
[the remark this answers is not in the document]
bensinjin: I tend to agree. I guess this does allow us to write the "we have two address types in the system" test. Maybe we can create a story to clean this stuff up based on a strategy we both agree on?

bc211/importer.py
```diff
@@ -120,3 +128,40 @@ def build_taxonomy_term_active_record(record, counters):
...
+        country=address_dto.country,
+        state_province=address_dto.state_province,
+        postal_code=address_dto.postal_code
+    )
+    if created:
+        counters.count_address()
+        LOGGER.info('Imported address: %s %s', active_record.id, active_record.address)
+    return active_record
+
+def create_address_type(address_type_id, counters):
```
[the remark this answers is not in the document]
bensinjin: The parser ensures we only parse what we're considering valid types. I kinda feel like the importer shouldn't be concerned with wether or not the address_type_id is valid.

bc211/importer.py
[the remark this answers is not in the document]
bensinjin: Probably not.  This does however allow us to test we actually imported two types but I your probably right, this may not be very useful and could be improved.

bc211/importer.py
```diff
@@ -120,3 +128,40 @@ def build_taxonomy_term_active_record(record, counters):
...
+        counters.count_address_type()
+        LOGGER.info('Imported address type: %s ', active_record.id)
+    return active_record
+
+def create_location_address(location, address, address_type):
+    active_record, created = LocationAddress.objects.get_or_create(
+        address=address,
+        location=location,
+        address_type=address_type
+    )
```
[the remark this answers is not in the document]
bensinjin: Yeah good point, fixed

bc211/parser.py
```diff
@@ -157,11 +159,60 @@ def is_bc211_taxonomy_term(code_str):
...
+    type_id = 'postal_address'
+    return parse_address(site.find('MailingAddress'), site_id, type_id)
+
+def parse_address(address, site_id, type_id):
+    address_lines = parse_address_lines(address)
+    city = parse_city(address)
+    country = parse_country(address)
+    if not address_lines or not city or not country:
+        LOGGER.debug('Unable to create address for location: %s. '
+                     'Parsed %s for address, %s for city, and %s for country.',
```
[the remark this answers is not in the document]
bensinjin: Fixed

bc211/tests/test_importer.py
bensinjin: Noted, however I think in this case it makes more sense to keep the test as is as all the importer tests read from one of the two files declared at the top of the file. Moving the contents of those files to somewhere more visible maybe doesn't make a whole lot of sense here?

bc211/importer.py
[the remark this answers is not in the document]
bensinjin: Address types are now created in a migration not from types parsed from the bc211 data.

bc211/tests/test_importer.py
bensinjin: Fixed

bc211/tests/test_parser.py
bensinjin: Fixed

bc211/tests/test_parser.py
bensinjin: Fixed

bc211/tests/test_parser.py
bensinjin: Fixed, we now get the location id from the test xml for these tests.

config/settings/base.py
bensinjin: Good to know

locations/models.py
bensinjin: See http://www.django-rest-framework.org/api-guide/relations/#reverse-relations. This allows us to properly embed the location address. We didn't need "address_locations" so I did removed that one. I've removed the on_delete on the address_type as I think it is a bit heavy handed probably.

locations/tests/helpers.py
bensinjin: Fixed. You're right we don't need this builder.

locations/tests/test_api.py
bensinjin: Sure, just a note, we actually no longer need to create the address type in this function as the types are automatically created in the db in the aforementioned migration.

locations/serializers.py
bensinjin: Probably? Really though I think we should discuss why we don't just have physical_address and postal_address fields directly on the location model instead of this addresses collection. I think that would be much simpler than addresses + address types.

bc211/tests/test_parser.py
bensinjin: Added better testing and test data as well as a warning when lines passed 4 are encountered.

addresses/admin.py
```diff
@@ -0,0 +1,6 @@
+from django.contrib import admin
+from addresses import models
+
+# Register your models here.
```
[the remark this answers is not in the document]
bensinjin: Fixed

addresses/migrations/0007_create_initial_address_types.py
```diff
@@ -0,0 +1,25 @@
+from django.db import migrations
+
+def forwards_func(apps, schema_editor):
+    AddressType = apps.get_model('addresses', 'AddressType')
+    db_alias = schema_editor.connection.alias
+    AddressType.objects.using(db_alias).bulk_create([
+        AddressType(id='physical_address'),
+        AddressType(id='postal_address')
+    ])
```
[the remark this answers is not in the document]
bensinjin: Sounds good

addresses/tests/test_models.py
```diff
@@ -0,0 +1,100 @@
...
+        address_from_db = validate_save_and_reload(address)
+        self.assertEqual(address_from_db.address, address.address)
+
+    def test_address_field_must_be_set_and_not_empty(self):
+        address = Address(city=a_string(), country=a_string(2, string.ascii_uppercase))
+        with self.assertRaises(exceptions.ValidationError):
+            address.save()
+        address = Address(city=a_string(), country=a_string(2, string.ascii_uppercase), address='')
+        with self.assertRaises(exceptions.ValidationError):
+            address.save()
```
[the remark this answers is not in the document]
bensinjin: Fixed

addresses/tests/test_models.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def test_has_city_field(self):
+        city_field_value = a_string()
+        address = AddressBuilder().with_city(city_field_value).build()
+        address_from_db = validate_save_and_reload(address)
+        self.assertEqual(address_from_db.city, address.city)
```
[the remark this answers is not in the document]
bensinjin: Fixed

addresses/tests/test_models.py
```diff
@@ -0,0 +1,100 @@
...
+    def test_city_field_must_be_set_and_not_empty(self):
+        address = Address(country=a_string(2, string.ascii_uppercase), address=a_string())
+        with self.assertRaises(exceptions.ValidationError):
+            address.save()
+        address = Address(city='', country=a_string(2, string.ascii_uppercase), address=a_string())
+        with self.assertRaises(exceptions.ValidationError):
+            address.save()
+
+    def test_has_country_field(self):
+        country_field_value = a_string(2, string.ascii_uppercase)
```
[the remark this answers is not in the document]
bensinjin: Fixed - I think I was durping on some these tests as my builder was already doing most of this work. Good catches.

bc211/importer.py
```diff
@@ -120,3 +124,31 @@ def build_taxonomy_term_active_record(record, counters):
...
+
+def create_address(address_dto, counters):
+    active_record, created = Address.objects.get_or_create(
+        address=address_dto.address_lines,
+        city=address_dto.city,
+        country=address_dto.country,
+        state_province=address_dto.state_province,
+        postal_code=address_dto.postal_code
+    )
```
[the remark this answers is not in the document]
bensinjin: Fixed

bc211/parser.py
```diff
@@ -157,11 +159,62 @@ def is_bc211_taxonomy_term(code_str):
...
+def parse_address(address, site_id, type_id):
+    address_lines = parse_address_lines(address)
+    city = parse_city(address)
+    country = parse_country(address)
+    if not address_lines or not city or not country:
+        LOGGER.warning('Unable to create address for location: "%s". '
+                       'Parsed "%s" for address, "%s" for city, and "%s" for country.',
+                       site_id, address_lines, city, country)
+        return None
+    address_type_id = type_id
```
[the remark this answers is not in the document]
bensinjin: Fixed

bc211/tests/test_parser.py
```diff
@@ -194,3 +225,50 @@ def test_sets_the_site_id(self):
...
+                                              self.location_id, 'physical_address')
+        self.assertEqual(parsed_address, None)
+
+    def test_none_is_returned_when_line1_is_empty(self):
+        address_lines = parser.parse_address_lines(self.physical_address)
+        self.assertEqual(address_lines, None)
+
+    def test_parsed_address_lines_correctly_formatted(self):
+        address_lines = parser.parse_address_lines(self.postal_address)
+        self.assertEqual(address_lines, 'Line1\nLine2\nLine3\nLine4')
```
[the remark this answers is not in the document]
bensinjin: Thanks for tips. I think I've improved the coverage and the tests in general (I needed to add more to actually cover the parsing of the entire address - not just address lines). I was under the impression at least having a Line1 in your address was required, everything else as far as lines are concerned would be optional is that not the case? Please peek at  252437be75403687a8b12887eab76e112ba73200 when you get a chance.

locations/models.py
```diff
@@ -43,6 +47,22 @@ def __str__(self):
...
+    address_type = models.ForeignKey(AddressType)
+
+    class Meta:
+        unique_together = ('location', 'address_type')
+
+    def __str__(self):
+        return '\"{address}\" for \"{location}\"'.format(
+            address=self.address,
+            location=self.location
+        )
```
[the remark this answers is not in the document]
bensinjin: Fixed

bc211/tests/test_parser.py
```diff
@@ -194,3 +230,124 @@ def test_sets_the_site_id(self):
...
+                <MailingAddress>
+                    <City>City</City>
+                    <Country>Country</Country>
+                </MailingAddress>
+            </Site>'''
+        root = etree.fromstring(xml_address)
+        address_type_id = 'postal_address'
+        with self.assertRaises(exceptions.MissingRequiredFieldXmlParseException):
+            parser.parse_address(root.find('MailingAddress'),
+                                 root.find('Key').text, address_type_id)
```
rasmus-storjohann-PG: Just one more change and then I am really happy with this PR. I can see that this test throws, but I am not confident that it is throwing for the right reason. So if you change the assert to check what the error message is on that exception, then I can be sure.  Here are my suggested changes:
```python
from bc211.exceptions import MissingRequiredFieldXmlParseException as MissingFieldException
from common.testhelpers.random_test_values import a_string
...
    def test_missing_line_1_throws_exception(self):
        # remove <Key> element
        xml_address = '''
            <Site>
                <MailingAddress>
                    <City>City</City>
                    <Country>Country</Country>
                </MailingAddress>
            </Site>'''
        root = etree.fromstring(xml_address)
        address_type_id = 'postal_address'
        # Specify what the error message should be. 
        # Use type alias for the exception class to make for a shorter line
        with self.assertRaisesRegex(MissingFieldException, r'Required field "Line1" not found'):
            # Pass in a_string() for key, to clearly show that the content of that string is irrelevant to this test
            parser.parse_address(root.find('MailingAddress'), a_string(), address_type_id)
````
The same patter shows up in several tests, so I think tidying this up is worth while.

bensinjin: Fixed

PR #183, opened 2018-02-07
WIP Feature/issue 106 user accounts

rasmus-storjohann-PG:


rasmus-storjohann-PG:
Closing this for now, after thinking about it some more I don't think getting user accounts in place is high priority at the moment.

Issue #154, opened 2017-12-21
Tidy Friday cleanup tasks: server


rasmus-storjohann-PG:
Task for tracking low effort tech debt cleanup tasks.

Tidy Friday PRs must meet the following criteria:

* Low developer effort
* No QA effort, correct by inspection

PR #188, opened 2018-02-09
Issue #154 marked some unit tests as slow

Lr-K:
FullDataImportTests are now marked as slow to exclude if desired (included by default).
To run unit tests ignoring slow process run:
python manage.py test --exclude-tag=slow

### DIFF
--- a/bc211/tests/test_importer.py
+++ b/bc211/tests/test_importer.py
@@ -2,6 +2,7 @@ import logging
 from bc211.importer import save_records_to_database
 from bc211.parser import read_records_from_file
 from django.test import TestCase
+from django.test import tag
 from human_services.locations.models import Location
 from human_services.organizations.models import Organization
 from human_services.taxonomies.models import TaxonomyTerm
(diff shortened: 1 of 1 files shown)


bc211/tests/test_importer.py
```diff
@@ -2,6 +2,7 @@
 from bc211.importer import save_records_to_database
 from bc211.parser import read_records_from_file
 from django.test import TestCase
+from django.test import tag
```
rasmus-storjohann-PG: This is not needed now that you don't use tags

bc211/tests/test_importer.py
```diff
@@ -97,32 +98,20 @@ def test_imports_correct_address_types(self):
...
+    
     def setUp(self):
         file = open(MULTI_AGENCY_FIXTURE, 'r')
         self.return_value = save_records_to_database(read_records_from_file(file))
         self.all_locations = Location.objects.all()
         self.all_organizations = Organization.objects.all()
         self.all_taxonomy_terms = TaxonomyTerm.objects.all()
 
-    def test_can_import_multiple_organizations(self):
+    def test_can_import_full_data_set(self):
```
rasmus-storjohann-PG: Could you do me a favour and stick a comment on here "breaking one-assert-per-test rule for faster test" or something like that.

Lr-K:
Done and pushed!

rasmus-storjohann-PG:
@bensinjin @robinpuga our first ever volunteer contributed PR!

PR #189, opened 2018-02-13
Feature/issue 177 setup swagger

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin this is work in progress, not yet ready for review. You may notice that it restructures the sorting code quite a bit, but I'll be happy to let you merge in the proximity sort stuff first.

bensinjin:
Cool sounds good.

rasmus-storjohann-PG:
If you go to https://fierce-ravine-89308-pr-189.herokuapp.com/swagger/ and make a call to /services/ with no argument, you'll see a 500 error. Can you have a look at the error message and let me know any insights you might have about what the problem could be?

bensinjin:
This is looking sensible. I've reviewed the commits but am not really as familiar with DRF as I should so I may find things to improve as I pull in my proximity search work.

PR #193, opened 2018-02-19
Feature/issue 192 fix screen output from importer

rasmus-storjohann-PG:


bensinjin:
This looks good, thanks for cleaning it up! I'm not sure why I assumed we would want to stop the importer if we couldn't parse an address, my bad.

PR #194, opened 2018-02-19
Feature/issue 186 search for multiple taxonomy terms

rasmus-storjohann-PG:


human_services/services/private.py
```diff
@@ -1,28 +1,31 @@
 from django.core.exceptions import SuspiciousOperation
 
 def parse_taxonomy_parameter(query_parameters):
-    taxonomy_term = query_parameters.get('taxonomy_term', None)
-    return build_valid_taxonomy_parameters(taxonomy_term)
+    parameter = query_parameters.get('taxonomy_term', None)
+    if not parameter:
+        return None
+    split_terms = [term.split(':') for term in parameter.split(',')]
```
bensinjin: Maybe break this up and be sure to call .strip().

human_services/services/private.py
```diff
@@ -1,28 +1,31 @@
...
+    return [build_valid_taxonomy_parameters(split_term) for split_term in split_terms]
 
 def build_valid_taxonomy_parameters(taxonomy_term):
-    if not taxonomy_term:
-        return None, None
-    if taxonomy_term.count(':') != 1:
+    if len(taxonomy_term) != 2:
```
bensinjin: 2 here is a bit magical, maybe a `taxomony_with_term_list_length = 2` variable would add clarity.

PR #195, opened 2018-02-22
Feature/issue 170 Proximity based sorting

bensinjin:
So here we have where I'm currently at with the proximity sorting piece. The initial commits are all around enforcing Postgres with PostGIS as our database layer and cleaning up configuration and documentation around using the .env file locally. There are a couple commits related to getting Travis playing nice with PostGIS that took some fiddling, it probably was not the best strategy to test if Travis builds pass by pushing to the branch but in this case getting it working wasn't too bad. Mixed in with this work I also migrate lat long float fields to Point objects, a requirement for our proximity searches. 

The latter half of the commits are around creating the filter required to sort by proximity. I've opted to create a base filter class called BaseProximityFilter and extend it for our pathways specific endpoints. I've taken this approach because our db layer in order to perform the distance query needs to know how to get a Point value from an object in a given queryset, I don't want to have to hardcode values specific to our ServiceAtLocation objects. You can see our specific ServiceAtLocation implementation here: https://github.com/pg-irc/pathways-backend/commit/4f1d3bc37b9847183b7105d88d428caeb5234ad2#diff-709ccfbc293eccc22a3f1981b8ed5322R78.

Anyways this is still a WIP just wanted to keep you up to speed.


bensinjin:
@rasmus-storjohann regarding subclassing the BaseProximityFilter. I've realized after learning more about how the filters work it wont be necessary to subclass this so we'll just have one filter. My initial assumption on having querysets of different object types was inaccurate.

human_services/locations/tests/test_api.py
```diff
@@ -110,17 +111,84 @@ def test_has_physical_address(self):
...
+                              .with_point(Point(49.246292, -123.116226)).create())
+        chilliwack_location = (LocationBuilder(self.organization)
+                              .with_name('Chilliwack')
+                              .with_point(Point(49.157940, -121.951467)).create())
+        kamloops_location = (LocationBuilder(self.organization)
+                              .with_name('Kamloops')
+                              .with_point(Point(50.674522, -120.327268)).create())
+        salmon_arm_location = (LocationBuilder(self.organization)
+                              .with_name('Salmon Arm')
+                              .with_point(Point(50.700103, -119.283844)).create())
```
rasmus-storjohann-PG: I like this test a lot. One thing that would make it easier to follow for someone not familiar with BC geography would be, probably jumble them up so that you can show that the sort is actually doing something and not just returning them in their order of creation, or you can use third_location as your target.
```
        first_location = (LocationBuilder(self.organization)
                              .with_name('First')
                              .with_point(Point(0, 0)).create())
        second_location = (LocationBuilder(self.organization)
                              .with_name('Second')
                              .with_point(Point(10, 10)).create())
        third_location = (LocationBuilder(self.organization)
                              .with_name('Third')
                              .with_point(Point(25, 25)).create())
        fourth_location = (LocationBuilder(self.organization)
                              .with_name('Fourth')
                              .with_point(Point(30, 30)).create())
```
I know, it's not as fun, but it is easier to follow, especially since I can look at the latlongs and be able to tell easily what the correct ordering ought to be

human_services/locations/tests/test_api.py
```diff
@@ -110,17 +111,84 @@ def test_has_physical_address(self):
...
+        ServiceLocationBuilder(self.service, vancouver_location).create()
+        ServiceLocationBuilder(self.service, chilliwack_location).create()
+        ServiceLocationBuilder(self.service, kamloops_location).create()
+        ServiceLocationBuilder(self.service, salmon_arm_location).create()
+        near_chilliwack_services_url = '{0}{1},{2}'.format('/v1/services_at_location/?proximity=',
+                                                           chilliwack_location.point.x,
+                                                           chilliwack_location.point.y)
```
rasmus-storjohann-PG: Picky, I think this is clearer
```
near_chilliwack_services_url = '/v1/services_at_location/?proximity={1},{2}'.format(
                                                           chilliwack_location.point.x,
                                                           chilliwack_location.point.y)
```

human_services/locations/viewsets.py
```diff
@@ -1,15 +1,24 @@
...
         return models.Location.objects.filter(organization=organization_id)
 
     serializer_class = serializers.LocationSerializer
+
+
+class ServiceAtLocationViewSet(viewsets.ReadOnlyModelViewSet):
+    queryset = models.ServiceAtLocation.objects.all()
+    search_fields = ('location__translations__name', 'service__translations__name')
+    serializer_class = serializers.ServiceAtLocationSerializer
+    filter_backends = (ServiceAtLocationProximityFilter,)
```
rasmus-storjohann-PG: Nice and simple

common/filters.py
```diff
@@ -53,15 +57,13 @@ def get_reference_point(self, request):
         if parameters:
             latitude_longitude_values = [parameter.strip() for parameter in parameters.split(',')]
             latitude_longitude_length = 2
-            if (len(latitude_longitude_values) is not latitude_longitude_length):
-                raise (ParseError('Exactly two comma separated values expected for {0}.'
-                                  .format(self.filter_parameter)))
+            if (len(latitude_longitude_values) != latitude_longitude_length):
+                raise (ParseError(self.parse_errors[0]))
```
rasmus-storjohann-PG: Hard to tell here whether `self.parse_errors[0]` is right or whether perhaps it should be `self.parse_errors[1]`. `self.parse_errors['should be two values']` or `self.parse_errors[TWO_VALUES_ERROR]` could work, or `self.two_values_parse_error`. Whatever works with the swagger library.

human_services/locations/documentation.py
```diff
@@ -0,0 +1,29 @@
...
+    operation_description = 'Get a list of services at locations'
+
+    manual_parameters = [openapi.Parameter('proximity',
+                                            openapi.IN_QUERY,
+                                            description=ServiceAtLocationProximityFilter.filter_description,
+                                            type=openapi.TYPE_STRING,
+                                            # Regex representing a latitude and longitude.
+                                            # Valid patterns consist of two integers or floats, comma separated, with
+                                            # optional + or - prefixes.
+                                            pattern=r'[-+]?[0-9]+\.?[0-9]*,[-+]?[0-9]+\.?[0-9]*'),
```
rasmus-storjohann-PG: You should permit white space around the comma, so `\W*,\W*`

human_services/locations/documentation.py
```diff
@@ -0,0 +1,29 @@
...
+                        ]
+
+    responses = {
+                    200: openapi.Response('A list of zero or more services at locations', ServiceAtLocationSerializer(many=True)),
+                    400: ', '.join(ServiceAtLocationProximityFilter.parse_errors),
```
rasmus-storjohann-PG: Nice

README.md
```diff
@@ -26,7 +26,17 @@ Install the required python libraries for local development
 pip install -r requirements/local.txt
 ```
 
-Create the database tables. For local development, sqlite is the database implementation, for production, postgres is used.
+Create an .env file for configuration overrides. See env.example for example use.
```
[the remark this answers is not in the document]
bensinjin: I've simplified this and added much clearer directions for setting up the local postgres environment.

README.md
```diff
@@ -26,7 +26,17 @@ Install the required python libraries for local development
...
+
+```
+touch .env
```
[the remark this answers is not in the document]
bensinjin: Fixed.

README.md
```diff
@@ -26,7 +26,17 @@ Install the required python libraries for local development
...
+```
+
+Create the database tables. Postgres with the PostGIS extension is required. Database default values are located in the
+config/settings/local.py, config/settings/test.py, and config/settings/production.py files. Overrides should be put in the .env file created in the previous step.
+
+To get up and running quickly just ensure you have a Postgres Superuser named "postgres" (default on fresh Postgres install)
```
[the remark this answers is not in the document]
bensinjin: I've updated the README to show how to create a user for local development so we can avoid using a superuser and simplified and shortened instructions.

config/settings/local.py
```diff
@@ -19,12 +19,9 @@
...
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-DATABASES = {
-    'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
-    }
-}
+DATABASES['default']['NAME'] = env('LOCAL_DATABASE', default='pathways_local')
+DATABASES['default']['USER'] = env('LOCAL_DATABASE_USER', default='postgres')
+DATABASES['default']['PASSWORD'] = env('LOCAL_DATABASE_USER', default='postgres')
```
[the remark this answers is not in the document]
bensinjin: Whoops, I've removed that password and the README now properly instructs on how to store your password in the .env file. This syntax is like this because now base.py defines the shared configuration across local.py, test.py, and production.py (like postgis as the ENGINE) and each individual file defines its environment's specific values on the DATABASES dictionary.

config/settings/test.py
```diff
@@ -6,15 +6,9 @@
...
-        'ENGINE': 'django.db.backends.postgresql',
-        'NAME': env('POSTGRES_DATABASE', default='test_db'),
-        'USER': env('POSTGRES_USER', default='postgres'),
-        'PASSWORD': env('POSTGRES_PASSWORD', default=''),
-        'ATOMIC_REQUESTS': True
-    }
-}
+DATABASES['default']['NAME'] = env('TEST_DATABASE', default='pathways_test')
+DATABASES['default']['USER'] = env('TEST_DATABASE_USER', default='postgres')
+DATABASES['default']['PASSWORD'] = env('TEST_DATABASE_USER', default='postgres')
```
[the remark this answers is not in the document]
bensinjin: Then ENGINE backend has changed (see previous note). I'm prefixing the databases with pathways to be consistant (Travis config now uses pathways_test) the default value for the password is a typo and has been removed.

env.example
```diff
@@ -1,7 +1,12 @@
...
+TEST_DATABASE_USER=test_user
+TEST_DATABASE_PASSWORD=test_password
+
+# Values used in local development
+LOCAL_DATABASE=local_database
+LOCAL_DATABASE_USER=local_user
+LOCAL_DATABASE_PASSWORD=local_password
 
 # Values used in production
 DATABASE_URL=postgres://user:password@host/dbname
```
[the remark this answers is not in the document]
bensinjin: Right, I'm actually not sure where the production example came from?

human_services/locations/migrations/0010_auto_20180208_1950.py
```diff
@@ -0,0 +1,29 @@
...
+        migrations.RemoveField(
+            model_name='location',
+            name='longitude',
+        ),
+        migrations.AddField(
+            model_name='location',
+            name='point',
+            field=django.contrib.gis.db.models.fields.PointField(blank=True, null=True, srid=4326),
+        ),
+    ]
```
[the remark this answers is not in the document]
bensinjin: Good point. The import will work after a rebase on master.

human_services/locations/tests/test_api.py
```diff
@@ -5,6 +5,8 @@
 from human_services.organizations.tests.helpers import OrganizationBuilder
 from human_services.addresses.tests.helpers import AddressBuilder
 from human_services.addresses.models import Address, AddressType
+from common.testhelpers.random_test_values import a_float
+from django.contrib.gis.geos import Point, GEOSGeometry
```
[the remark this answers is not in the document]
bensinjin: Fixed.

human_services/locations/tests/test_api.py
```diff
@@ -107,3 +109,18 @@ def test_has_physical_address(self):
 
     def test_has_postal_address(self):
         self.location_has_address_of_type('postal_address')
+
+    def test_has_point(self):
+        point = Point(a_float(), a_float())
+        LocationBuilder(self.organization).with_point(point).create()
+        url = '/v1/locations/'
+        response = self.client.get(url)
+        self.assertEqual(response.json()[0]['point'], point.__str__())
```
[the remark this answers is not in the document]
bensinjin: Fixed in later commits.

common/filters.py
```diff
@@ -32,3 +36,45 @@ class SearchFilter(filters.SearchFilter):
...
                           'currently only looks in name and description, make it look more '
                           'widely.')
+
+
+class BaseProximityFilter(filters.BaseFilterBackend):
+    filter_parameter = 'proximity'
+    filter_description = ('Order by proximity to a point. '
+                          'Accepts two comma values representing a latitude and a longitude. '
+                          'For example: proximity=+49.2827,-123.1207.')
+    srid = 4326
```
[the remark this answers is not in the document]
bensinjin: I think in this instance srid has more meaning than "Spatial Reference System Identifier" as it's more widely used in the mapping space. I wouldn't generally use abbreviations but I think this may be an exception.

common/filters.py
```diff
@@ -32,3 +36,45 @@ class SearchFilter(filters.SearchFilter):
...
+
+    def get_model_point_field(self):
+        raise NotImplementedError('.get_model_point_field() must be overridden.')
```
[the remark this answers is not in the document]
bensinjin: Cheers, this is in the process of being fixed (I felt the same way right after pushing it off)

common/filters.py
```diff
@@ -32,3 +36,45 @@ class SearchFilter(filters.SearchFilter):
...
+                raise (ParseError('Exactly two comma separated values expected for {0}.'
+                                  .format(self.filter_parameter)))
+            try:
+                latitude = float(latitude_longitude_values[0])
+                longitude = float(latitude_longitude_values[1])
+            except ValueError:
+                raise (ParseError('Values provide to {0} must be able to represent integers or floats.'
+                                 .format(self.filter_parameter)))
+            return Point(latitude, longitude, srid=self.srid)
+        return None
```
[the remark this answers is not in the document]
bensinjin: Yep, agreed, WIP

human_services/locations/tests/test_api.py
```diff
@@ -110,17 +111,84 @@ def test_has_physical_address(self):
...
+        longitude = float(response.json()[0]['longitude'])
+        Point(latitude, longitude)
+
+class ServiceAtLocationProximityFilterTests(rest_test.APITestCase):
+    def setUp(self):
+        self.organization = OrganizationBuilder().create()
+        self.service = ServiceBuilder(self.organization).create()
+
+    def test_200_response_when_proximity_includes_plus(self):
+        url = '/v1/services_at_location/?proximity=+11.1111,+222.2222'
```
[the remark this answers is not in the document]
bensinjin: Will do!

.travis.yml
```diff
@@ -9,14 +9,14 @@ addons:
...
-  - DATABASE_URL: postgres://postgres@localhost/test_db
+  - DATABASE_URL: postgres://postgres@localhost/pathways_test
   - DJANGO_AWS_STORAGE_BUCKET_NAME: peacegeeks-pathways-static
   DJANGO_SECRET_KEY:
     secure: EBC8CfYNMc/N4EZof8raivIzqgbYRQSAXEvIDMYQIId6CKqASW2IVFn+zxXLJcL7m/D+jmgWPlu81mcvZKekin8zhRfx6IKXqgr2TzahvzM9pMtLClcMHABcmwan63i3AexCEC0FHnufVb17QqQjwjeIy78zfsU8GK38+UjuuyIrd/0u2XWo6XPCfoK2a74gBR1bCnJd5UnBmD45XiQHBgbYt+mHZEnSZs6o8TVGpa5Y5ovaRqhlY2Ydy1avTHLtm9KdrioXAHvmDEx8g6YuVmfu7WWvC5UBFDo3o4SAwSpq1IMXm8s39EZ37i1x/UWI/20CPM2nWtGMcBz/9ihm8xVSDMeZQAIx0Hkz/g6+9SCXzyaubL+fiqZpcv6xiYbPsYEVFLZRGimneCHYaouPCEU8JPj4BdFYWdS0RtYjd7rhC9CIXnZJ1mHaPpa7dI39/UtZd5+POo3YNOXUgdBqvOjlQc7VsVbDeQxvs9nKU1c5DWV5tWNdvDnuQ5Dxr5WrAKX1zwY4JQ5nMIGXcv1I3y5rgtP9Xsi6oMp5NQe0uTUUVj4YB15EdLBBi1/UFvswDbBv8QmJMzay0IMGrPbV8REUWYYaPrC4Ylg+FSp2LbIuCBqWDQ3X/79X4FOB1KQentHBLCSo8+kG3d0Bjjx+SAGIqxKZ7V0Ozhh299ZNHJE=
 install:
 - pip install -r requirements/test.txt
 before_script:
-- psql -c 'create database test_db;' -U postgres
+- psql -c 'create database pathways_test;' -U postgres
```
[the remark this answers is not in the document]
bensinjin: Yeah ok, I'm reverting back to old defaults.

human_services/locations/tests/test_api.py
bensinjin: Good point, this has been fixed.

human_services/locations/tests/test_api.py
bensinjin: Fixed.

human_services/locations/viewsets.py
bensinjin: :) agreed.

human_services/locations/documentation.py
bensinjin: Fixed

human_services/locations/tests/test_api.py
```diff
@@ -107,3 +109,18 @@ def test_has_physical_address(self):
...
+        url = '/v1/locations/'
+        response = self.client.get(url)
+        self.assertEqual(response.json()[0]['point'], point.__str__())
+
+    def test_point_value_instantiates_to_geos_geometry(self):
+        LocationBuilder(self.organization).create()
+        url = '/v1/locations/'
+        response = self.client.get(url)
+        # This will throw one of a few various exceptions
+        # if parsing the value fails
```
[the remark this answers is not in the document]
bensinjin: This comment was only here to communicate that the test framework will cause a failure when an exception is thrown. This wasn't immediately obvious to me (I thought you had to assert a specific exception was thrown) so I wanted to make it clear to others.

common/filters.py
bensinjin: Yeah originally this was devised with Swagger in mind. I am working on improving this though as I'm not a huge fan of the magical index that represents the error we're looking for. WIP

bensinjin:
@rasmus-storjohann This has just been rebased (and some commits squashed) and needs a review. 
Would you also be able to actually checkout this branch locally and take a look at the swagger endpoints? I'm noticing (which you're probably already aware of) we're getting additional "read" endpoints generated for us, for example /locations/{location_id}/services/{service_id}/services_at_location/{id}/ which is actually akin to how you described /locations/{location_id}/services/{service_id}/services_at_location/ working in the issue. I personally think services_at_location endpoints should be list endpoints as DRF assumes.

Finally, I've not yet added tests for translations when using location.name and service.name in serializers (see ServiceAtLocationSerializer). We don't yet have the infrastructure in place to make requests based on language which is another Issue which I feel should probably encompass testing our serializer's ability to return results in the correct language.

README.md
```diff
@@ -26,7 +26,23 @@ Install the required python libraries for local development
...
+```
+$psql postgres
+postgres=# CREATE USER pathways WITH PASSWORD 'your-secure-password';
+postgres=# CREATE DATABASE pathways_local;
+postgres=# GRANT ALL PRIVILEGES ON DATABASE pathways_local TO pathways;
+postgres=# \connect pathways_local
+postgres=# CREATE EXTENSION postgis;
+```
+
+Create a .env file and add your database user password
```
rasmus-storjohann-PG: Add "Did you know that if you put a space before the shell command, the command is not saved in the shell history. Handy with commands that contain confidential information:"

config/settings/local.py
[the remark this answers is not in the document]
rasmus-storjohann-PG: I like the updates to the README.

env.example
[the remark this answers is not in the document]
rasmus-storjohann-PG: I'm hoping that people can get started developing without worrying about .env, and when they do need it, the example file that is there now gives people a pretty good idea of how it should be used. However, the example .evn file containing values for production is actually misleading, so if you can just remove those lines from the example file, then I think we're good.

human_services/locations/tests/test_api.py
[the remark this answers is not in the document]
rasmus-storjohann-PG: Yeah, an uncaught exception in a test means that the test fails, I think that will be common knowledge.

common/filters.py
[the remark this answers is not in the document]
rasmus-storjohann-PG: OK

human_services/locations/documentation.py
rasmus-storjohann-PG: Need to figure out if `\s` or `\W` is the right thing to use for white space, currently two regexs are inconsistent.

common/filters.py
```diff
@@ -32,3 +39,78 @@ class SearchFilter(filters.SearchFilter):
...
+
+class ProximityFilter(filters.BaseFilterBackend):
+    filter_description = ('Order by proximity to a point. '
+                          'Accepts two comma separated values representing a latitude and a longitude. '
+                          'Example: "proximity=+49.2827,-123.1207".')
+    srid = 4326
+
+    def filter_queryset(self, request, queryset, view):
+        proximity_parameter = request.query_params.get('proximity', None)
+        if proximity_parameter:
```
rasmus-storjohann-PG: Early return on None will make for shallower nesting and cleaner code.

common/filters.py
```diff
@@ -32,3 +39,78 @@ class SearchFilter(filters.SearchFilter):
...
+            proximity = ProximityParameterParser(proximity_parameter).parse()
+            if queryset.model is ServiceAtLocation:
```
rasmus-storjohann-PG: Should this class no-op or throw if it is used on the wrong model? Same question applies to all the filters here.

common/filters.py
```diff
@@ -32,3 +39,78 @@ class SearchFilter(filters.SearchFilter):
...
+            taxonomy_terms = TaxonomyParameterParser(taxonomy_parameter).parse()
+            for term in taxonomy_terms:
+                if queryset.model is Service:
+                    queryset = queryset.filter(taxonomy_terms__taxonomy_id=term[0],
+                                               taxonomy_terms__name=term[1])
+                elif queryset.model is ServiceAtLocation:
+                    queryset = queryset.filter(service__taxonomy_terms__taxonomy_id=term[0],
+                                               service__taxonomy_terms__name=term[1])
+
+        return queryset
```
rasmus-storjohann-PG: This function is too long, you can break it up. Early return will help too.

common/filters.py
```diff
@@ -32,3 +39,78 @@ class SearchFilter(filters.SearchFilter):
...
+
+class OrganizationIdFilter(filters.BaseFilterBackend):
+    filter_description = ('Filter by the organization id path parameter.')
+
+    def filter_queryset(self, request, queryset, view):
+        organization_id = view.kwargs.get('organization_id', None)
+        if organization_id:
+            if queryset.model is Service:
+                queryset = queryset.filter(organization_id=organization_id)
+        return queryset
```
rasmus-storjohann-PG: I really like this change. Filters is clearly the way to go.

common/tests/test_filter_parameter_parsers.py
```diff
@@ -0,0 +1,47 @@
...
+from common.filter_parameter_parsers import ProximityParameterParser, TaxonomyParameterParser
+
+class ProximityParameterParserTests(TestCase):
+    def test_parses_proximity_with_white_space(self):
+        proximity = ProximityParameterParser(' 1.1111, 22.2222    ').parse()
+        self.assertEqual(proximity[0], 1.1111)
+        self.assertEqual(proximity[1], 22.2222)
+
+    def test_throws_when_proximity_has_less_than_two_values(self):
+        with self.assertRaisesRegex(ParseError, ProximityParameterParser.errors['exactly_two_values']):
```
rasmus-storjohann-PG: I really like these tests. However, I don't think the tests should reference the `errors` objects. That is really an implementational detail, and we should be able to change how the parsers organize their error strings without changing the tests. The tests shouldn't even care about all of the error string, it should just have enough in the test to verify that the correct error is returned, so in this case a regex on `two comma separated values expected` is enough.

config/urls.py
```diff
@@ -5,23 +5,29 @@
 from django.views.generic import TemplateView
 from django.views import defaults as default_views
 from human_services.organizations.viewsets import OrganizationViewSet
-from human_services.locations.viewsets import LocationViewSet, LocationViewSetUnderOrganizations
+from human_services.locations.viewsets import (LocationViewSet, LocationViewSetUnderOrganizations,
+                                               ServiceAtLocationViewSet)
```
rasmus-storjohann-PG: We'll be able to remove `LocationViewSetUnderOrganizations` but we can do that later.

human_services/locations/documentation.py
```diff
@@ -0,0 +1,41 @@
...
+
+    manual_parameters = [openapi.Parameter('proximity',
+                                            openapi.IN_QUERY,
+                                            description=ProximityFilter.filter_description,
+                                            type=openapi.TYPE_STRING,
+                                            # Regex representing a latitude and longitude.
+                                            # Valid patterns consist of two integers or floats, comma separated, with
+                                            # optional + or - prefixes. One space after the comma is allowed here.
+                                            pattern=r'^[-+]?[0-9]+\.?[0-9]*,\s?[-+]?[0-9]+\.?[0-9]*$'),
+                         # TODO come up with a strategy to not duplicate from documentation in services app
```
rasmus-storjohann-PG: A file `common/documentation.py` with one or more functions in it that returns what you need here?

human_services/locations/documentation.py
```diff
@@ -0,0 +1,41 @@
...
+                                            # where a term consits of one or more letters/dashes, a dot, and more
+                                            # letters/dashes. White space is allowed around the comma but not the
+                                            # dots
+                                            pattern=r'^[\w\-]+\.[\w\-]+(\W*,\W*[\w\-]+\.[\w\-]+)*$'),
+                        ]
+
+    responses = {
+                    200: openapi.Response('A list of zero or more services at locations', ServiceAtLocationSerializer(many=True)),
+                    400: (', '.join([ProximityParameterParser.errors_to_string(),
+                          TaxonomyParameterParser.errors_to_string()])),
```
rasmus-storjohann-PG: I think I would have had `ProximityParameterParser.errors_to_string()` and `TaxonomyParameterParser.errors_to_string()` both return arrays of strings, and then combine the arrays here and do the join once. It would allow you the flexibility of e.g. sorting the errors alphabetically before joining them, remove duplicates, stuff like that.

human_services/locations/tests/helpers.py
```diff
@@ -18,12 +17,8 @@ def with_name(self, name):
...
         return self
 
-    def with_latitude(self, latitude):
-        self.latitude = latitude
-        return self
-
-    def with_longitude(self, longitude):
-        self.longitude = longitude
+    def with_point(self, point):
+        self.point = point
```
rasmus-storjohann-PG: Your tests might become a little bit simpler if you had `def with_point(self, lat, long)`, so you don't have to call the Point constructor everywhere.

human_services/locations/tests/helpers.py
```diff
@@ -45,6 +39,7 @@ def create(self):
         result.save()
         return result
 
+
 class ServiceLocationBuilder:
```
rasmus-storjohann-PG: You can actually get rid of this class. Since `ServiceAtLocation` doesn't have any fields that can be set to arbitrary values, you will always specify all fields, so the builder pattern doesn't buy you anything over just calling the `ServiceAtLocation` constructor directly.

human_services/locations/tests/test_api.py
```diff
@@ -109,3 +111,149 @@ def test_has_physical_address(self):
...
+        self.assertEqual(response.json()[0]['latitude'], point.x)
+        self.assertEqual(response.json()[0]['longitude'], point.y)
+
+    def test_point_values_instantiate_a_point(self):
+        LocationBuilder(self.organization).create()
+        url = '/v1/locations/'
+        response = self.client.get(url)
+        latitude = float(response.json()[0]['latitude'])
+        longitude = float(response.json()[0]['longitude'])
+        Point(latitude, longitude)
```
rasmus-storjohann-PG: What does this test actually test? That the Point constructor doesn't throw?

human_services/locations/tests/test_api.py
```diff
@@ -109,3 +111,149 @@ def test_has_physical_address(self):
...
+        json = response.json()
+        self.assertEqual(json[0]['location_name'], second_location.name)
+        self.assertEqual(json[1]['location_name'], first_location.name)
+        self.assertEqual(json[2]['location_name'], third_location.name)
+        self.assertEqual(json[3]['location_name'], fourth_location.name)
+
+    def test_can_full_text_search(self):
+        service_at_locations = self.create_two_service_at_locations()
+        expected_service_at_location = service_at_locations[1]
+        response = self.client.get('/v1/services_at_location/?search=sec')
```
rasmus-storjohann-PG: If you do a full text search using the name or description from `expected_service_at_location` as search term this test will make more sense. Then you can also remove the hard-coded strings in `create_two_service_at_locations()`.

human_services/locations/tests/test_api.py
```diff
@@ -109,3 +111,149 @@ def test_has_physical_address(self):
...
+    def test_can_filter_by_location_id(self):
+        service_at_locations = self.create_two_service_at_locations()
+        expected_service_at_location = service_at_locations[0]
+        location_id = expected_service_at_location.location.id
+        response = (self.client.get('/v1/locations/{0}/services_at_location/'
+                                    .format(location_id)))
+        json = response.json()
+        self.assertEqual(len(json), 1)
+        self.assertEqual(json[0]['location_name'], expected_service_at_location.location.name)
+        self.assertEqual(json[0]['service_name'], expected_service_at_location.service.name)
```
rasmus-storjohann-PG: Nice, simple tests. Beautiful.

human_services/locations/tests/test_api.py
```diff
@@ -109,3 +111,149 @@ def test_has_physical_address(self):
...
+
+    def test_can_page(self):
+        service_at_locations = self.create_two_service_at_locations()
+        expected_service_at_location = service_at_locations[1]
+        response = self.client.get('/v1/services_at_location/?per_page=1&page=2')
+        json = response.json()
+        self.assertEqual(len(json), 1)
+        self.assertEqual(json[0]['location_name'], expected_service_at_location.location.name)
+        self.assertEqual(json[0]['service_name'], expected_service_at_location.service.name)
```
rasmus-storjohann-PG: Not sure that paging is working unless you have more matching records in the database than what can fit in a page, and you can show that only a page worth of records is returned. See `createMany()` idea below.

human_services/locations/tests/test_api.py
```diff
@@ -109,3 +111,149 @@ def test_has_physical_address(self):
...
+
+    def test_can_filter_by_taxonomy(self):
+        taxonomy_terms = [TaxonomyTermBuilder().create(), TaxonomyTermBuilder().create()]
```
rasmus-storjohann-PG: When you're doing something non-trivial with a builder, think about whether you can enhance the builder to do what you need, so that it'll be easier in the future. Here you could do
```
taxonomy_terms = TaxonomyTermBuilder().createMany(2)
```
(At Aquatic, we had this exact thing, and the number of elements defaulted to three.)

human_services/locations/tests/test_models.py
```diff
@@ -62,42 +64,16 @@ def test_name_cannot_be_none(self):
         with self.assertRaises(django_utils.IntegrityError):
             validate_save_and_reload(location)
 
-    def test_has_latitude(self):
-        latitude = 123.456
-        location = LocationBuilder(self.organization).with_latitude(latitude).build()
+    def test_can_set_point(self):
+        point = Point(a_float(), a_float())
```
rasmus-storjohann-PG: `a_point()`

human_services/locations/tests/test_models.py
```diff
@@ -62,42 +64,16 @@ def test_name_cannot_be_none(self):
...
+        location = LocationBuilder(self.organization).with_point(point).build()
         location_from_db = validate_save_and_reload(location)
-        self.assertAlmostEqual(location_from_db.latitude, latitude)
+        self.assertAlmostEqual(location_from_db.point, point)
```
rasmus-storjohann-PG: How does this work?

common/filters.py
```diff
@@ -32,3 +39,78 @@ class SearchFilter(filters.SearchFilter):
                           'by space or comma. Logical AND is implied among the terms. TODO '
                           'currently only looks in name and description, make it look more '
                           'widely.')
+
+
+class ProximityFilter(filters.BaseFilterBackend):
+    filter_description = ('Order by proximity to a point. '
+                          'Accepts two comma separated values representing a latitude and a longitude. '
+                          'Example: "proximity=+49.2827,-123.1207".')
```
rasmus-storjohann-PG: The example looks funny in swagger, since the user should not enter the proximity part. Better to go with `'Example: "+49.2827,-123.1207".'`.

.travis.yml
```diff
@@ -1,11 +1,15 @@
...
+addons:
+  postgresql: 9.6
+  apt:
+    packages:
+    - postgresql-9.6-postgis-2.3
 env:
   global:
   - DJANGO_SETTINGS_MODULE: config.settings.test
-  - DATABASE_URL: postgres://postgres@localhost/test_db
+  - DATABASE_URL: postgres://postgres@localhost/pathways_test
```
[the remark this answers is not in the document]
bensinjin: This was reverted later

common/filter_parameter_parsers.py
```diff
@@ -0,0 +1,81 @@
+from rest_framework.exceptions import ParseError
```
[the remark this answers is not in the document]
bensinjin: That's not entirely true. Documentation references errors and calls methods in these. Is private.py a Django convention? I understand the idea that a file called private.py would be for private functionality but I don't like that it likely eventually becomes a module you just dump things if you don't know where to put them...

common/filter_parameter_parsers.py
```diff
@@ -0,0 +1,81 @@
...
+
+class ProximityParameterParser:
+    errors = {
+        'exactly_two_values': 'Exactly two comma separated values expected for proximity',
+        'invalid_latitude_value_type': 'Latitude value provided to proximity must be able to represent a float',
+        'invalid_longitude_value_type': 'Longitude value provided to proximity must be able to represent a float',
+    }
+
+    def __init__(self, parameter_value):
+        self.parameter_value = parameter_value
```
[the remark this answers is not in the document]
bensinjin: Cool, yeah I tend to agree. I've moved the parsing work to constructors for both the parameter parsers in this file.

common/filter_parameter_parsers.py
```diff
@@ -0,0 +1,81 @@
...
+    def build_valid_longitude_value(self, value):
+        try:
+            valid_value = float(value)
+        except ValueError:
+            raise (ParseError(self.errors['invalid_longitude_value_type']))
+        return valid_value
+
+    def parse(self):
+        if not self.parameter_value:
+            return None
```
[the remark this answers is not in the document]
bensinjin: Moving this work to constructor (done) ensures we no longer get invalid parameter data.

common/filter_parameter_parsers.py
```diff
@@ -0,0 +1,81 @@
...
+        split_terms = [term.split('.') for term in self.parameter_value.split(',')]
+        return [self.build_valid_taxonomy_pair(split_term) for split_term in split_terms]
+
+    @classmethod
+    def errors_to_string(cls):
+        return ('{0}, {1}, {2}'.format(cls.errors['exactly_two_values'],
+                                cls.errors['empty_taxonomy_id'],
+                                cls.errors['empty_taxonomy_term']))
```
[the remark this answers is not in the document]
bensinjin: This is actually a dictionary so join wont work here, however, this does:
```python
', '.join(list(self.errors.values()))
```

common/filters.py
```diff
@@ -32,3 +39,78 @@ class SearchFilter(filters.SearchFilter):
...
+    srid = 4326
```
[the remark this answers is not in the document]
bensinjin: I've moved this to settings.

common/filters.py
bensinjin: I would tend to agree in most instances but this feels slightly different. If we think in terms of a set of conditions passing in order for a filter to be applied I think this is easier to follow.  I think reading this as "If we have a none empty proximity parameter and we're acting on a collection of ServiceAtLocation objects then apply the filter" is clearer than "If we have an empty proximity parameter, return the unfiltered list, if not see if we're acting on a collection of ServiceAtLocation objects, if so filter, if not return the unfiltered list". For my filters I've actually referenced how DRF implements its filters and I found their approach to very easy to follow so I've adopted it here.

common/filters.py
bensinjin: I don't think that's necessary although I did consider it. My assumption here is that if you need these filters on new viewsets your going to come in here and add a case for collections of your viewset's object type. If we run into problems with this approach I don't think it will be much work later to tweak this implementation...

common/filters.py
bensinjin: This ties into two comments back. 9 lines is too long? I personally like that fact that if possible our filters can do their work in filter_queryset alone although I'm open to refactored approach if you have one? This particular filter was basically just copied from our old "privates.py" in the services app that used to do this work there.

common/tests/test_filter_parameter_parsers.py
bensinjin: Good point, fixed

human_services/locations/documentation.py
bensinjin: Yeah ok. You can see my solution at common/documentation.py

human_services/locations/documentation.py
bensinjin: Good thinking, fixed

human_services/locations/tests/helpers.py
bensinjin: Fixed

human_services/locations/tests/test_models.py
bensinjin: Fixed

human_services/locations/tests/helpers.py
bensinjin: Good call, removed and ported existing tests to just use ServiceAtLocation objects

human_services/locations/tests/test_api.py
bensinjin: That's a bit of a nutty test and I've actually opted to remove it. It was basically meant to validate lats and longs provided by the api could be validated as floats which is definitely not clear here. The thing is this requirement is already tested in the parser tests, also, it would be impossible for our api to return values incompatible with a Point as values returned by the serializer are already properties of a Point object.

human_services/locations/tests/test_api.py
bensinjin: Fixed

human_services/locations/tests/test_api.py
bensinjin: In this instance paging is indeed working. The `expected_service_at_location` is checking for the second (index 1) service at location object created here that we'd expect to see on page 2 of we were limiting this to 1 item per page. I do agree however that create_many is a good idea for bulk creation in builders.

human_services/locations/tests/test_api.py
bensinjin: Cool. I've added this to TaxonomyTermBuilder and am using it here.

human_services/locations/tests/test_models.py
bensinjin: Fixed

human_services/locations/tests/test_models.py
bensinjin: Not very clear at all. I've cleaned this up to just call valid

common/filters.py
bensinjin: Fixed

bensinjin:
@rasmus-storjohann I've addressed your comments here. Re. "Where are the tests that show that the API throws on sorting by both proximity and other fields?" - I don't think these tests are necessary as our viewset explicitly defines what filters it accepts and sort_by is not one of them ..

common/documentation.py
```diff
@@ -0,0 +1,26 @@
+from drf_yasg import openapi
+from common.filters import ProximityFilter, TaxonomyFilter
+
+class ManualParameters:
```
rasmus-storjohann-PG: A class with just static members is functionally equivalent to a namespace or module. I think that's all you need here. (A namespace is simpler than a class, and always use the simplest tool at your disposal.)

common/documentation.py
```diff
@@ -0,0 +1,26 @@
...
+    @staticmethod
+    def get_proximity_parameter():
+        return (openapi.Parameter('proximity',
+                                   openapi.IN_QUERY,
+                                   description=ProximityFilter.filter_description,
```
rasmus-storjohann-PG: Makes sense that the filter description also live in this file then, since it's definitely 'documentation'.

human_services/locations/tests/test_api.py
```diff
@@ -119,13 +119,6 @@ def test_has_point_values(self):
         self.assertEqual(response.json()[0]['latitude'], location.point.x)
         self.assertEqual(response.json()[0]['longitude'], location.point.y)
 
-    def test_point_values_instantiate_a_point(self):
```
rasmus-storjohann-PG: Fix grammar in commit message, `git rebase -i`

human_services/locations/tests/test_api.py
rasmus-storjohann-PG: Yeah, the test will be easier to follow if the numbers are in their common range, so with create_many (or a dedicated helper function as part of the test class) to create maybe 30 entities, and then get page 2 with 10 per page.

human_services/taxonomies/tests/helpers.py
```diff
@@ -24,3 +24,9 @@ def create(self):
         result = self.build()
         result.save()
         return result
+
+    def create_many(self, n=3):
+        terms = []
+        for _ in range(0, n):
+            terms.append(TaxonomyTermBuilder().create())
```
rasmus-storjohann-PG: I think you can do this in one line with list comprehension.

human_services/locations/tests/test_models.py
```diff
@@ -68,8 +68,7 @@ def test_name_cannot_be_none(self):
 
     def test_can_set_point(self):
         location = LocationBuilder(self.organization).with_point(a_float(), a_float()).build()
-        location_from_db = validate_save_and_reload(location)
-        self.assertIsInstance(location_from_db.point, Point)
+        validate_save_and_reload(location)
```
rasmus-storjohann-PG: Tests should assert something, otherwise I'm not sure why it's here.

common/tests/test_filter_parameter_parsers.py
```diff
@@ -9,39 +9,39 @@ def test_parses_proximity_with_white_space(self):
         self.assertEqual(proximity.longitude, 22.2222)
 
     def test_throws_when_proximity_has_less_than_two_values(self):
-        with self.assertRaisesRegex(ParseError, ProximityParser.errors['exactly_two_values']):
+        with self.assertRaises(ParseError):
```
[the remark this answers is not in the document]
bensinjin: Right, fixed

common/documentation.py
bensinjin: The class adds meaning - this is a class that houses manual parameter definitions for Swagger. The module name "documentation" is already very vague and it's conceivable we'll be using it for  other documentation type functionality.. thoughts?

common/documentation.py
bensinjin: This is the description of what our drf filter does and is independent from our Swagger parameter documentation it just so happens we've opted it's currently sufficient for both

human_services/locations/tests/test_api.py
bensinjin: Fixed

common/documentation.py
rasmus-storjohann-PG: I think the function names within a file called 'documentation' will communicate what the functions are about. It's true that we may add more documentation to this file, but we can do that without a class as long as we name the functions well.

human_services/taxonomies/tests/helpers.py
bensinjin: Fixed

common/documentation.py
bensinjin: Fixed

common/documentation.py
bensinjin: Ok, moved for now but I think we should maybe do some Swagger research and see if we can't make swagger just pull the descriptions from our filters automatically. I think it already may do this for other filters provided by drf (see https://github.com/encode/django-rest-framework/blob/master/rest_framework/filters.py) by inspecting the "search_description" and "filter_description" properties? Could be wrong ...

human_services/locations/tests/test_models.py
bensinjin: I've added clarity around this test .. what's wrong with just testing that something doesn't throw? I don't think you can assert that but a test will fail if it does.

bensinjin:
Think this is getting ready to ship

common/filter_parameter_parsers.py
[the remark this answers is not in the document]
rasmus-storjohann-PG: private.py is not a Django convention, it's something I picked up at Aquatic. They are in a sense grab-bags of various things, but as long as our apps are small I(which most of them are), there should not be that much in each private.py. I would be open to `private-foo.py` and `private-bar.py` files, but I would want to make sure that the need for those files are not due to the module actually getting too big and the proper solution is to break the module (or app) up into smaller pieces. On that note, I would like to see the ServieAtLocation broken out into its own app at one point.

common/filters.py
rasmus-storjohann-PG: It's not the number of lines as such, but the amount of logic that is making this function too long. One for loop and to if statements is a lot, have a look at what Uncle Bob says about functions doing one thing.

human_services/locations/tests/test_models.py
rasmus-storjohann-PG: It's a bit unusual but can be valid. I would make sure the name of the test made it clear that the assert wasn't left off by mistake, such as `foo_does_not_throw()`.

rasmus-storjohann-PG:
I'm going to have another look over it tomorrow and merge it then when my head is clearer.

human_services/locations/tests/test_models.py
bensinjin: Ah good point. I've since tweaked this test to check we can properly save a point object on a location and that the point on the location is what we'd expect

bensinjin:
@rasmus-storjohann I've just rebased off master and would like you to take one more look. I have cleaned up the documentation bugs you recently noticed as well as adding the appropriate documentation for the "page" parameter. We should probably add an issue around ensuring all our endpoint have the appropriate documentation including potential 400 errors.  Your other suggestions can probably be discussed/moved to their own issues if that's ok?

PR #196, opened 2018-03-01
Feature/no issue wagtail exploration

rasmus-storjohann-PG:
@bensinjin and @robinpuga for your perusal.


bensinjin:
This seems sensible. I think our focus has shifted a bit to the front end. For now I'm just going to leave this here.

rasmus-storjohann-PG:
The reason I put this up now is that I'm thinking ahead to ask AppScoop to help with the server side, and I wanted to get this change in before they get started so that they can stay completely out of the human_services part of the system.

Issue #170, opened 2018-01-15
Service search: sort by proximity


rasmus-storjohann-PG:
See main story #163

We assume that the client needs information about services, ranked by the services' respective locations' distance to a reference point provided in the query. There is a many2many relationship between services and locations, meaning that there are many proximity values for a given service. Also, the client is probably interested in information about both the services and the locations. It follows that the return type from this end point should not be an array of Services, but rather a array of objects that contain information from both a service and a location. This leads naturally to a distinct service end point supporting proximity search.

`/services_at_location/` supports proximity sorting as well as all the same filter functionality as `/services/`. However, sorting by proximity and some other field using `sort_by` in the same request is not supported: It does not make sense to sort both by proximity and any other field, since the probability that any two locations will have exactly the same proximity value is low. (Except when multiple services are offered at the same location, hmmm.) This API returns a list of ServiceAtLocation objects that contains data from both the service and location records. 

ACs:
* `/services_at_location/?proximity=-123.456,+23.45`, returns an array of ServiceAtLocation with the service name and location name in the default language
* Support filtering consistent with the `/services/` endpoint: `/services_at_location/?proximity=-123.456,+23.45&search=foobar&taxonomy=foo.bar&page=2&per_page=25`. This will probably involve porting the existing taxonomy work to be a filter
* `/locations/{id}/services_at_location/` returns all the services at a given location, sorting by proximity doesn't make sense in this context, but is it invalid? It's similar to `/locations/{id}/services/` but except that it returns ServiceAtLocation.
* `/services/{id}/services_at_location/?proximity=-123.456,+23.45` returns all the locations for a given service, here sorting by proximity does make sense. It is similar to `/services/{id}/locations/`, except that it returns ServiceAtLocation and supports proximity sort.
* `/locations/{id}/services/{id}/services_at_location/` and `/services/{id}/locations/{id}/services_at_location/` return information about a given service at a given location. It returns one object of type ServiceAtLocation, not an array. It is similar to `/locations/{id}/services/{id}` and `/services/{id}/locations/{id}` except for the return type.

bensinjin:
@rasmus-storjohann Can we just do /services?proximity=49.2827,-123.1207  as the format for these urls? Not sure I understand the $?

bensinjin:
@rasmus-storjohann Please read my essay:

I think we might want to consider changing how we do our proximity filtering for services. 

[...]

PR #199, opened 2018-03-14
Issue #170 Create PostGIS integration on Heroku

bensinjin:
This along with the Heroku python and geo buildpacks enable our PostGIS functionality on Heroku.  I'm adding a script to run setup tasks (for instance migrate) as this was not currently in place and was causing the 500 errors.  There appears from what I can tell to be an issue where python is not able to pickup GDAL and GEOS library environment variables from Heroku during the deploy phase and this is why I've added production_heroku.py with these values. I have tried setting these values in the app.json as well as the UI and it almost feels like possibly python's virtual environment cannot see them?? I'm not 100% but really it's probably not a huge deal but will continue to be on my radar.

Another note - I think we should be importing the prod db (or something sensible) into our review app's dbs so we can fully utilize this feature.  Currently the review app db is empty as you'll notice. I've checked this branch out here: https://fierce-ravine-89308.herokuapp.com/swagger/ so you can see it with data against the stage db.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
-release: ./release-tasks.sh
+release: python manage.py migrate
 web: gunicorn config.wsgi:application --log-file -
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
+release: ./release-tasks.sh
 web: gunicorn config.wsgi:application --log-file -
--- a/app.json
+++ b/app.json
@@ -38,7 +38,8 @@
   ],
   "buildpacks": [
     {
-      "url": "heroku/python"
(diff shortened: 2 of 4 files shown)


Procfile
```diff
@@ -1 +1,2 @@
+release: ./release-tasks.sh
```
rasmus-storjohann-PG: Would this work?
```
release: python manage.py migrate
```

bensinjin: yep - I think we'll want to add to this down the road though.

bensinjin: Ok I've pushed a fix to remove this for the time being.

PR #201, opened 2018-03-24
Feature/issue 154 services at location

rasmus-storjohann-PG:


bensinjin:
Good stuff. I think it's nice to break this out.

Issue #204, opened 2018-05-31
Taxonomy refactor


rasmus-storjohann-PG:
Taxonomy terms will be associated with newcomers guide data as well as human services data. It's location in the src folder structure needs to reflect this. Also, change property name "name" to "id" for consistency.

rasmus-storjohann-PG:
Done

PR #205, opened 2018-05-31
Issue #204 Move taxonomy module so that is can be easily accesses by …

rasmus-storjohann-PG:
…both human_services and newcomers' guide modules


rasmus-storjohann-PG:
@bensinjin , can you review this very simple change on the backend? It's a simple reorganization in anticipation that the HSDS taxonomy table will be needed in newcomers' guide data too, so it's location in the folder structure should reflect that.

Issue #114, opened 2017-10-25
Design data model for representing entities with text fields in multiple languages


rasmus-storjohann-PG:
Data model should support translating text fields in arbitrary entities to an arbitrary set of languages. It supports knowing when data was added or altered, requiring new translations, and it supports knowing when translations were done. When a field is altered, it supports knowing what the previous value was, to help the translators.

The data model is compatible with transifex formats for each extraction of raw data for translation and easy ingestion of completed translations.

PR #210, opened 2018-07-04
Issue pg-irc/pathways-frontend#114 Add questionnaire taxonomies to th…

rasmus-storjohann-PG:
…e system taxonomy map


rasmus-storjohann-PG:
@bensinjin for review, mop up of a few outstanding things server side for the import and recommendation engine to work

Issue #217, opened 2018-08-19
Newcomers Guide import: Fix links with hyphens in the domain name


trisapeace:
AC:
* URLs that have hyphens in their domain names (for example `http://www.vast-vancouver.ca`) will generate the such that they are clickable within the app.

Currently in `articles.js`, I see 
`[www.vast](http://www.vast)-vancouver.ca`

when it should be 
`[www.vast-vancouver.ca](http://www.vast-vancouver.ca)`

trisapeace:
@rasmus-storjohann-PG I think that this one is important to fix.

PR #222, opened 2018-08-21
Issue #217 Fix import bug for URLs containing a dash in the host name

rasmus-storjohann-PG:


bensinjin:
@rasmus-storjohann this looks good. We'll get #223 in first then this.

bensinjin:
@rasmus-storjohann fix these conflicts and we'll move onto #223

PR #227, opened 2018-08-24
Feature/issue 226 fix whitespace to protect list formatting

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin and @trisapeace for review, Theresa you might want to have a look at the new unit tests, since they give an idea of how we need to format the content. There should be minimal changes from how things have been done so far.

Issue #218, opened 2018-08-19
Newcomers Guide import: Fix titles with periods


trisapeace:
AC:
* Article and Task titles that have periods in them (for example `B.C. Health`) are fully displayed in the app

Currently, it appears that it is truncating the name at the first period, so instead of seeing 
`B.C. Health`

the article appears in the app to be named
`B`
which is not useful.

trisapeace:
@rasmus-storjohann-PG I think that this one is important to fix.

PR #232, opened 2018-08-28
Issue #218 Fix bug where import script would throw on valid non-conte…

rasmus-storjohann-PG:
…nt files, e.g. taxonomy.txt


rasmus-storjohann-PG:
@bensinjin this PR fixes a bug introduced last week

Issue #226, opened 2018-08-22
Newcomers' Guide import: fix bullet and numbered lists


rasmus-storjohann-PG:
Need to tweak the importer cleanup code to produce text formatted to match the mark-down library's behaviour around bullet lists and numbered lists. It's tricky because there is a bug https://github.com/mientjan/react-native-markdown-renderer/issues/74 that means we need to remove newline characters generally, but we need to leave them unchanged around lists.

PR #233, opened 2018-08-28
Issue #226 Tweak newline handling

rasmus-storjohann-PG:
With this change, we avoid a lot of cleanup in the content, specifically converting `\n\n` into `\n\n\n` and `\n` into `\n\n`.


rasmus-storjohann-PG:
@bensinjin for review

PR #234, opened 2018-08-28
Feature/issue frontend 217 Populate fixture with service queries from the content files

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin you might want to have a look at this one two, it's the server piece to go with the client serviceQuery work you looked at yesterday. Forgot to flag you on this one yesterday.

Issue #241, opened 2018-09-11
Add additional search fields for services_at_location endpoint.


bensinjin:
AC: Full text searches at the services_at_location endpoint search against the service title, service description, location title, and location description

PR #242, opened 2018-09-11
Issue #241 Add additional fields to search on at services_at_location

bensinjin:
The services endpoint searches across these fields we should also be doing so at the services_at_location endpoint.


human_services/services_at_location/viewsets.py
```diff
@@ -11,7 +11,8 @@
 class ServiceAtLocationViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = models.ServiceAtLocation.objects.all()
     serializer_class = serializers.ServiceAtLocationSerializer
-    search_fields = ('location__translations__name', 'service__translations__name')
+    search_fields = ('location__translations__name', 'location__translations__description',
+                     'service__translations__name', 'service__translations__description')
```
rasmus-storjohann-PG: There should be tests for this

bensinjin: Fixed

PR #256, opened 2018-10-15
Issue #264 (frontend) Remove article fixture generation and related a…

bensinjin:
…rticles

### DIFF
--- a/newcomers_guide/read_data.py
+++ b/newcomers_guide/read_data.py
@@ -7,19 +7,9 @@ def read_task_data(root_folder):
     for root, _, filenames in os.walk(root_folder, topdown=False):
         for filename in filenames:
             path = os.path.join(root, filename)
-            if is_task_file(path):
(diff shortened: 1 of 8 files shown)


newcomers_guide/read_data.py
```diff
@@ -7,21 +7,12 @@ def read_task_data(root_folder):
     for root, _, filenames in os.walk(root_folder, topdown=False):
         for filename in filenames:
             path = os.path.join(root, filename)
-            if is_task_file(path):
+            if is_task_file(path) or is_article_file(path):
+                print(path)
```
rasmus-storjohann-PG: The output from this tool is important in our content manangement work flow (such as it is), so I'd rather not have all the files printed out like this. Can you also change the commit message to contain "Issue pg-irc/pathways-frontend#264", that way it's clickable in github which is better.

bensinjin: Whoops, that wasn't meant to make it in. Removed.

PR #258, opened 2018-10-16
Issue #257 Replace \u2028 newline characters with \n

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review, this came up when processing the Arabic content.

PR #261, opened 2018-10-26
Feature/issue 251 loosen address validation

bensinjin:


bc211/parser.py
```diff
@@ -202,67 +202,46 @@ def parse_airs_taxonomy_term(code_str):
...
-
-
-def parse_postal_address(site_xml, site_id, site_name):
-    type_id = 'postal_address'
-    return parse_address_and_handle_errors(site_xml.find('MailingAddress'), site_id, site_name, type_id)
+def parse_physical_address(site_xml, site_id):
+    physical_address = site_xml.find('PhysicalAddress')
+    if physical_address:
+        return parse_address(physical_address, site_id, 'physical_address')
+    return None
```
rasmus-storjohann-PG: It's preferred to return early on error, not success, so this should be
```py
if not physical_address:
    return None
return parse...
```
It gets a lot cleaner when there are multiple error conditions that cause early return, and a nice thing to be consistent with. So if you like you can do something like this:
```py
def parse_address(address, site_id, address_type_id):
    city = parse_city(address)
    if not city:
        return None
    country = parse_country(address)
    if not country:
        return None
```

It applies a few times in this PR.

bc211/parser.py
```diff
@@ -202,67 +202,46 @@ def parse_airs_taxonomy_term(code_str):
...
-
-    state_province = parse_state_province(address)
-    postal_code = parse_postal_code(address)
-    return dtos.Address(location_id=site_id, address_lines=address_lines,
-                        city=city, state_province=state_province, postal_code=postal_code,
-                        country=country, address_type_id=address_type_id)
+    if city and country:
+        return dtos.Address(location_id=site_id, address_type_id=address_type_id, city=city, country=country,
+                            address_lines=parse_address_lines(address), state_province=parse_state_province(address),
+                            postal_code=parse_postal_code(address))
```
rasmus-storjohann-PG: My IDE complains about long lines here

bc211/parser.py
```diff
@@ -202,67 +202,46 @@ def parse_airs_taxonomy_term(code_str):
...
-        if re.match("Line[2-4]", child.tag):
-            address_line = parse_required_field(address, child.tag)
-            address_lines.append(address_line)
-        if re.match("Line[5-9]", child.tag):
-            LOGGER.warning('Tag %s encountered and has not been parsed.', child.tag)
-    return '\n'.join(address_lines)
+    sorted_address_children = sorted(address.getchildren(), key=lambda child: child.tag)
+    address_line_tags = [child.tag for child in sorted_address_children if re.match('Line[1-9]', child.tag)]
+    address_line_tag_values = [parse_optional_field(address, tag) for tag in address_line_tags]
+    none_empty_address_lines = [address_line for address_line in address_line_tag_values if address_line]
```
rasmus-storjohann-PG: spelling: 'non_empty_address_lines', otherwise really nice cleanup

bc211/tests/test_dtos.py
```diff
@@ -163,3 +163,33 @@ def test_throws_on_missing_phone_number_type_id(self):
...
+        with self.assertRaises(exceptions.MissingRequiredFieldXmlParseException):
+            dtos.Address(location_id=a_string(), city=a_string(), country=a_string())
+
+    def test_throws_on_missing_city(self):
+        with self.assertRaises(exceptions.MissingRequiredFieldXmlParseException):
+            dtos.Address(location_id=a_string(), address_type_id=a_string(), country=a_string())
+
+    def test_throws_on_missing_country(self):
+        with self.assertRaises(exceptions.MissingRequiredFieldXmlParseException):
+            dtos.Address(location_id=a_string(), address_type_id=a_string(), city=a_string())
```
rasmus-storjohann-PG: Thanks, this was an oversight! 

How do you feel about having tests like `test_does_not_throw_on_missing_xyz()` as well? Do you think those tests would be of benefit?

bc211/tests/test_parser.py
```diff
@@ -276,42 +276,41 @@ def test_missing_line_1_throws_exception(self):
...
+        address = parser.parse_address(root.find('MailingAddress'), site_id, address_type_id)
+        self.assertIsInstance(address, dtos.Address)
 
-    def test_missing_city_throws_exception(self):
+    def test_fails_silently_on_empty_city(self):
         xml_address = '''
             <Site>
                 <MailingAddress>
                     <Line1>Line1</Line1>
+                    <City />
```
rasmus-storjohann-PG: It should work the same whether the City element is absent or empty, right?

bc211/parser.py
bensinjin: Fixed (in all places)

bc211/parser.py
bensinjin: Fixed

bc211/parser.py
bensinjin: Fixed

bc211/tests/test_dtos.py
bensinjin: Yeah, I like this, tests added.

bc211/tests/test_parser.py
bensinjin: Good catch, added absent tests.

bensinjin:
@rasmus-storjohann Changes addressed

Issue #259, opened 2018-10-17
Ensure topic ids are unique


rasmus-storjohann-PG:
We have a topic in chapters 3 and 9 that are both called "Where to go for help". We are using these titles as ids in the data model. It's a problem that they're not unique.

It's a closely related problem for the UX that the titles are ambiguous, the better solution is to rename these sections to be more meaningful. Should still add validation to the importer to ensure that task ids are unique.

PR #262, opened 2018-10-27
Issue #259 Throw on the same task id being used in different chapters

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

PR #263, opened 2018-10-31
Issue pg-irc/pathways-frontend#279 Change links to appear as just [li…

rasmus-storjohann-PG:
…nk] and [email]


rasmus-storjohann-PG:
@bensinjin for review

PR #266, opened 2018-11-07
Feature/issue 265 nlp based search

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review. There's quite a lot here, so we can discuss it tomorrow and I can bring you up to speed on it.

search/migrations/0005_auto_20181109_1739.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.2 on 2018-11-09 17:39
```
bensinjin: You'll likely want to delete your migrations files along with their associated tables and re run `makemigrations` so we only get one migration per table in this PR.

rasmus-storjohann-PG: That is a bit of a dangerous road to go down, it will make it harder to migrate data later. See https://docs.djangoproject.com/en/2.1/topics/migrations/ The migration files are generated files and as such it is not really a readability issue for the codebase as a whole whether there are 1 or 4 files in there.

bensinjin: Here you are introducing new tables. There is no data to be migrated so you don't need to worry about that and a paper trail of your tweaks isn't that useful. Am I missing something? Squashing migrations looks to me to be a clean up task that you *may* do at some point.  Either way, how are you feeling about this PR? Are you still working on it?

rasmus-storjohann-PG:
Let's aim to merge this PR early next week. I don't want to upset things before tomorrow (although it shouldn't do anything, I don't want to be wrong about that just now). I will also use this PR to iron out any minor Heroku issues.

PR #276, opened 2018-11-28
Feature/issue frontend 301 update bc211 data on heroku

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

PR #279, opened 2018-12-06
Feature/frontend 315 show nlp data

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review even with the failing test, I'll fix that when I address your comments

config/urls.py
```diff
@@ -33,6 +34,7 @@ def build_router():
     router.register(r'locations/(?P<location_id>\w+)/services/(?P<service_id>\w+)/services_at_location',
                     ServiceAtLocationViewSet)
     router.register(r'services_at_location', ServiceAtLocationViewSet, base_name='services_at_location')
+    router.register(r'tasks/(?P<task_id>\w+)/related_tasks', RelatedTasksViewSet, base_name='tasks')
```
bensinjin: Is there a reason we don't see this endpoint here: https://fierce-ravine-89308-pr-279.herokuapp.com/v1/?

bensinjin: Same for task services.

rasmus-storjohann-PG: Digging into this

rasmus-storjohann-PG: Hmmm, the API route response has a lot of different issues with it, many that I am sure go back in time. We don't currently depend on the content of this response, so I am fine with living with this as is until we have need to fix it.

bensinjin: Ok.

PR #282, opened 2018-12-14
Issue #154 Add command to create client fixture to prepare deploy script

rasmus-storjohann-PG:


rasmus-storjohann-PG:
@bensinjin for review

PR #283, opened 2018-12-19
Issue #154 Fix problem where data dump files caused duplicate primary…

rasmus-storjohann-PG:
… key error on load


rasmus-storjohann-PG:
@bensinjin for review and merge if it's OK

PR #286, opened 2018-12-21
Feature/issue 285 make server report its version

rasmus-storjohann-PG:


rasmus-storjohann-PG:
Need to document the REST endpoint

main/__init__.py
```diff
@@ -1,2 +1,4 @@
-__version__ = '0.1.0'
-__version_info__ = tuple([int(num) if num.isdigit() else num for num in __version__.replace('-', '.', 1).split('.')])
+__version__ = '1.0.0'
+
+__version_info__ = tuple([int(num) if num.isdigit() else num for num in
```
bensinjin: Can you elaborate on the dash replacement here?

rasmus-storjohann-PG: So this code is actually from the Django cookiecutter bootstrap, but it makes sense. I just changed the formatting (VS code probably did it automatically). So version strings like "1.0.0" and "1.0-0" etc all become [1, 0, 0] at a tuple. I don't expect we'll deviate from the dotted form.

bensinjin: Gotcha, thx

PR #298, opened 2019-01-18
Feature/issue 297 server performance

rasmus-storjohann-PG:


rasmus-storjohann-PG:
Tested server from app, seems to work OK

PR #307, opened 2019-02-20
Issue #464 adjust headings in newcomers guide

tomy-pg:


rasmus-storjohann-PG:
Closing because this PR doesn't actually solve the problem, and we have no alternative solution at the moment.

PR #311, opened 2019-03-01
Issue #466 remove HTML from bc211 content responses

tomy-pg:


bc211/parser.py
```diff
@@ -288,12 +291,13 @@ def convert_phone_type_to_type_id(phone_type):
 
 
 def remove_double_escaped_html_markup(data):
-    unescaped_once = html.unescape(data)
-    unescaped_twice = html.unescape(unescaped_once)
+    if data: 
+        unescaped_once = html.unescape(data)
+        unescaped_twice = html.unescape(unescaped_once)
```
rasmus-storjohann-PG: There are good reasons for returning early (or throwing early) on error. It's more readily apparent when you have more than one thing that could go wrong. Compare
```py
if ok:
    if ok:
        if ok:
            return valid result
         else:
            return error 3
      else:
          return error 2
   else:
      return error 1
```
I hope you agree that this is better, less indents, the return error bit is close in the code to the corresponding test
```py
if not ok:
    return error 1
if not ok:
    return error 2
if not ok:
    return error 3
return valid result
```

bc211/tests/test_parser.py
```diff
@@ -458,3 +459,55 @@ def build_phone_xml(self, phone_number, phone_number_type):
...
+        xml_id = '''
+        <Source>
+            <Agency>
+                <Key>&amp;lt;b&amp;gt;abc</Key>
+            </Agency>
+        </Source>
+        '''
+        root = etree.fromstring(xml_id)
+        self.html_markup = parser.parse_agency_key(root.find('Agency'))
+        self.assertEqual(self.html_markup, 'abc')
```
rasmus-storjohann-PG: Don't need the `self.` prefix on `html_markup`, you can use local variables rather than member variables

bc211/tests/test_parser.py
```diff
@@ -458,3 +459,55 @@ def build_phone_xml(self, phone_number, phone_number_type):
                     <Type>{}</Type>
                 </Phone>
             </Site>'''.format(phone_number, phone_number_type)
+
+
+class HTMLMarkupParserTests(unittest.TestCase):
+    def test_removes_doubly_escaped_bold_markup_from_required_field(self): 
+        xml_id = '''
```
rasmus-storjohann-PG: Why the name `xml_id`?

search/compute_similarities.py
```diff
@@ -21,44 +19,10 @@ def to_service_ids_and_descriptions(services):
     descriptions = []
     for service in services:
         ids.append(service.id)
-        description_without_markup = remove_double_escaped_html_markup(service.description)
-        descriptions.append(service.name + ' ' + description_without_markup)
```
rasmus-storjohann-PG: I suggest you combine this commit with the one where you added `remove_double_escaped_html_markup` to a different part of the code, that will make it clear that this code was moved. You can do this using an interactive rebase

search/tests/test_compute_similarities.py
```diff
@@ -55,45 +55,6 @@ def test_getting_description_for_service_returns_name_and_description(self):
...
-        name = 'Representation Services'
-        description_with_markup = ('in situations such as getting an immediate court order to '
-                                   'ensure client&#39;s or their children&#39;s safety and security')
-        expected_without_markup = ('Representation Services '
-                                   'in situations such as getting an immediate court order to '
-                                   'ensure client\'s or their children\'s safety and security')
-        ServiceBuilder(self.organization).with_name(name).with_description(description_with_markup).create()
-        _, descriptions = to_service_ids_and_descriptions(Service.objects.all())
-        self.assertEqual(descriptions[0], expected_without_markup)
-
```
rasmus-storjohann-PG: It's a very good rule to live by that all unit tests should pass at every commit. Sometimes it's hard to live by that rule, but it's pretty rare. These tests fail after you removed the tested function in the previous commit.

bc211/parser.py
```diff
@@ -291,13 +291,15 @@ def convert_phone_type_to_type_id(phone_type):
...
+
+    unescaped_once = html.unescape(data)
+    unescaped_twice = html.unescape(unescaped_once)
 
-        remover = HTMLRemover()
-        remover.feed(unescaped_twice)
-        return remover.get_data()
+    remover = HTMLRemover()
+    remover.feed(unescaped_twice)
+    return remover.get_data()
```
rasmus-storjohann-PG: This change is an improvement

bc211/parser.py
```diff
@@ -43,8 +43,8 @@ def parse_agency_key(agency):
 
 def parse_required_field(parent, field):
     try:
-        text_of_field = parent.find(field).text
-        return remove_double_escaped_html_markup(text_of_field)
+        field_text = parent.find(field).text
+        return remove_double_escaped_html_markup(field_text)
     except AttributeError:
         raise MissingRequiredFieldXmlParseException('Missing required field: "{0}"'.format(field))
```
rasmus-storjohann-PG: Now with enhanced field names! ;-) 

Alternative commit message is "Rename fields"

bc211/parser.py
```diff
@@ -40,14 +43,18 @@ def parse_agency_key(agency):
...
 
 
 def parse_optional_field(parent, field):
     value = parent.find(field)
-    return None if value is None else value.text
+    if value is None:
+        return None
+    field_text = value.text
+    return remove_double_escaped_html_markup(field_text)
```
rasmus-storjohann-PG: `return remove_double_escaped_html_markup(value.text)` would be fine here, but you don't need to change it.

bc211/parser.py
rasmus-storjohann-PG: I reviewed each commit, some of my comments don't make a lot of sense when looking at all the changes together

rasmus-storjohann-PG:
From the BC211 data set:
```
The &amp;lt;b&amp;gt;Resettlement Assistance Program&amp;lt;/b&amp;gt; offers settlement and immigration services for government-assisted refugees.
```
This same passage is rendered in the app as
```
The &lt;b&gt;Resettlement Assistance Program&lt;/b&gt; offers settlement and immigration services for government-assisted refugees.
```

PR #312, opened 2019-03-01
Issue #473 Modify rightaway icon to check-square-o

DarwishMenna:


DarwishMenna:
@rasmus-storjohann-PG

PR #313, opened 2019-03-01
Issue pg-irc/pathways-frontend#501 Improve screen output for Newcomer…

rasmus-storjohann-PG:
…s Guide import script


rasmus-storjohann-PG:
@bensinjin this should be ready for merge, it's a couple of simple changes to improve screen output from the import script

Issue #314, opened 2019-03-05
URLs with trailing `=` parameters are truncated


rasmus-storjohann-PG:
![image](https://user-images.githubusercontent.com/32307781/53823780-4840ee00-3f27-11e9-97b5-569ed6ce9b99.png)

rasmus-storjohann-PG:
Expected is that urls like this http://foo.com=434 are included in the link, observed is that the trailing =434 is left out.

PR #316, opened 2019-03-06
Issue #314 fix urls with trailing equals

tomy-pg:
I also addressed urls ending with '/' from being truncated as well.

### DIFF
--- a/newcomers_guide/clean_data.py
+++ b/newcomers_guide/clean_data.py
@@ -101,7 +101,7 @@ def unprotect_newlines(text):
 
 
 def clean_up_http_links(text):
-    return re.sub(r'(https?://[^\s/][^\s]*[a-zA-Z]*[=/]?[a-zA-Z0-9/?=]+)', r'[link](\1)', text)
+    return re.sub(r'(https?://[^\s/][^\s]*[a-zA-Z0-9/])', r'[link](\1)', text)
     
 
 def clean_up_email_links(text):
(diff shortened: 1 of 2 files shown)


newcomers_guide/tests/test_clean_data.py
```diff
@@ -269,6 +269,22 @@ def test_excludes_trailing_closing_parenthesis(self):
...
+        self.assertEqual(clean_up_http_links(text), 'abc [link](http://example.com/) def')
+
+    def test_http_link_does_not_truncate_query(self):
+        text = 'abc http://example.com/search?source=abc def'
+        self.assertEqual(clean_up_http_links(text), 'abc [link](http://example.com/search?source=abc) def')
+    
+    def test_http_link_does_not_truncate_query_with_ampersand(self):
+        text = 'abc http://example.com/search?source=a&page=b def'
+        self.assertEqual(clean_up_http_links(text), 'abc [link](http://example.com/search?source=a&page=b) def')
+
```
rasmus-storjohann-PG: Additional test cases:
'abc http://example.com/search?source=a&page=b. def'
'abc http://example.com/search?source=a&page=b, def'
'abc http://example.com/search def'

newcomers_guide/clean_data.py
```diff
@@ -101,8 +101,8 @@ def unprotect_newlines(text):
 
 
 def clean_up_http_links(text):
-    return re.sub(r'(https?://[^\s/][^\s]*[a-zA-Z])', r'[link](\1)', text)
-
```
rasmus-storjohann-PG: return re.sub(r'(https?://[^\s/][^\s]*[a-zA-Z])([,\.]\s)', r'[link](\1)\2', text)

PR #320, opened 2019-03-07
Issue #504 remove phone numbers before recommendations are made

tomy-pg:


search/tests/test_compute_similarities.py
```diff
@@ -62,3 +62,57 @@ def test_computing_similarity_matrix(self):
...
         self.assertLess(similarity_matrix[0, 2], 0.10)
+
+    def test_removes_local_phone_numbers_from_description(self):
+        name = 'Government of Canada'
+        description_with_phone_numbers = 'Call 778-123-4567 or 604-123-4567 for more information.'
+        description_without_phone_numbers = ('Government of Canada ' 
+                                             'Call  or  for more information.')
+        ServiceBuilder(self.organization).with_name(name).with_description(description_with_phone_numbers).create()
+        _, descriptions = to_service_ids_and_descriptions(Service.objects.all())
+        self.assertEqual(descriptions[0], description_without_phone_numbers)
```
rasmus-storjohann-PG: This is a nice test. You can simplify it a bit by omitting the name, so `ServiceBuilder` will just put in a random string, but then you can change the assertion to use `assertIn` to just match the part of the string you care about. It will make your test more expressive since you're clearly stating what it is you care about in the return value.

tomy-pg: Ah yes, I see. This would certainly make it more clear what I'm focused on testing here.

search/compute_similarities.py
```diff
@@ -36,3 +38,6 @@ def compute_similarities(docs):
 def compute_cosine_doc_similarities(matrix):
     normalized_matrix = sklearn.preprocessing.normalize(matrix, axis=1)
     return normalized_matrix * normalized_matrix.T
+
+def remove_phone_numbers(description):
+    return re.sub(r'(\+?[0-9]{1,2}-|\+?[\(\)\d]{3,5}-\d{3}-\d{4})', '', description)
```
bensinjin: I personally think commenting regular expressions is useful (dunno if @rasmus-storjohann  agrees). Knowing your intention here can make tracking down potential bugs later easier.

rasmus-storjohann-PG: Yeah, regexes sometimes warrant comments (I tend to discourage comments elsewhere in code). In this case the unit test are also a good source to understand this regex, I think.

tomy-pg: I think the unit tests I have in place help make this regex understandable but if its confusing other people I am certainly happy to comment.

PR #331, opened 2019-03-19
Issue #552 Add organization name to service serializer

DarwishMenna:


rasmus-storjohann-PG:
Looks good, will merge when freeze is over

PR #343, opened 2019-04-09
Pin down Django to 2.1.7

cotsog:


rasmus-storjohann-PG:
Solution used in a different PR. Thanks @cotsog !

Issue #352, opened 2019-04-23
BC211 import script should return None when provided non-countries.


tomy-pg:
There are five services that have <Country>All Countries</Canada> in their Mailing/Physical Address. 

AC:
- When a non-country is provided, it should return None.

tomy-pg:
Another option is to return N/A. We would need to increase the max character length to 3. Doing NA alone would be misleading as NA is short for Namibia.

rasmus-storjohann-PG:
QA: Checked data imported from source with <Country>All countries</Country> for their mailing address. All these record had valid, complete physical addresses. All affected mailing addresses where dropped on import, as required.

PR #354, opened 2019-04-25
Issue #352 fix bc211 import script to return None when non-country is provided

tomy-pg:
Also in this PR is a fix for Python 3.7. I was receiving this when running tests “RuntimeError: generator raised StopIteration”. 

More details here: https://docs.python.org/3/whatsnew/3.7.html

### DIFF
--- a/bc211/tests/test_parser.py
+++ b/bc211/tests/test_parser.py
@@ -242,7 +242,7 @@ class AddressParserTests(unittest.TestCase):
                 <PhysicalAddress>
                     <Line1>Line1</Line1>
                     <City>City</City>
-                    <Country>Country</Country>
+                    <Country>Canada</Country>
                 </PhysicalAddress>
             </Site>'''
         root = etree.fromstring(xml_address)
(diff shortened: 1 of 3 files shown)


tomy-pg:
It is not clear to me why three tests were not failing before. These tests had <Country>Country</Country> and were being returned as 'Country' even though we have countries set to max 2 characters. 

For now I replaced 'Country' with 'Canada' in these three tests.

bc211/tests/test_parser.py
```diff
@@ -308,6 +308,19 @@ def test_replaces_united_states_with_corresponding_two_letter_code(self):
         address = parser.parse_address(root.find('PhysicalAddress'), a_string(), 'physical_address')
         self.assertEqual(address.country, 'US')
 
+    def test_fails_silently_on_non_country(self):
```
rasmus-storjohann-PG: test_sets_address_to_none_on_invalid_country

bc211/tests/test_parser.py
```diff
@@ -308,6 +308,19 @@ def test_replaces_united_states_with_corresponding_two_letter_code(self):
         address = parser.parse_address(root.find('PhysicalAddress'), a_string(), 'physical_address')
         self.assertEqual(address.country, 'US')
 
+    def test_sets_address_to_none_on_invalid_country(self):
```
rasmus-storjohann-PG: Check to see if the names of the other tests can be improved by following this pattern

Issue #345, opened 2019-04-03
Server returns bc211 last database update date


jparlina:
People often asked how often the Arrival Advisor gets updated data from bc211.
Could we show "Last Update from bc211 database: ______" on the About page of the Arrival Advisor app where we show the app and server version?

rasmus-storjohann-PG:
ACs: * Similar to how the server returns it's version at endpoint /version, return the database version at /dbversion

rasmus-storjohann-PG:
QA: Confirmed that version string for data set is returned from end point `https://fierce-ravine-89308.herokuapp.com/bc211version/`

PR #355, opened 2019-04-29
Issue #345 server returns last bc211 update date 

tomy-pg:


BC211_LAST_UPDATE.txt
```diff
@@ -0,0 +1 @@
+20-03-2019
```
rasmus-storjohann-PG: Remove

bc211/version.py
```diff
@@ -0,0 +1,17 @@
+# The version is defined in /pathways-backend/bc211/__init__.py
```
rasmus-storjohann-PG: Regarding the file name, if you look at existing filenames, you'll find that they are named after the kinds of classes they contain. So this file should be called views.py.

tomy-pg: Fixed. I was following the existing file names in /main when I named it version.py.

rasmus-storjohann-PG: Yeah, that file name is not very consistent. /main is a bit of a grab bag folder, unfortunately.

Issue #317, opened 2019-03-07
Fix intermittend test failure


rasmus-storjohann-PG:
```
Traceback (most recent call last):
  File "/home/rasmus/git/pathways-backend/search/tests/test_save_similarities.py", line 35, in test_saves_required_number_of_records_for_each_row
    self.assertEqual(TaskSimilarityScore.objects.count(), scores_saved_in_all)
AssertionError: 11 != 10

```

rasmus-storjohann-PG:
Test picks 25 random integers. The test will fail if two of these happen to be the same. Probably not something to worry about.

tomy-pg:
Fixing this because this recently caused Travis to not pass my PR #214

PR #359, opened 2019-05-06
Issue #317 fix intermittent test failure

tomy-pg:
Fixed the sometimes failing test `test_saves_required_number_of_records_for_each_row` in both `TestSavingTaskSimilarities` and `TestSavingTaskServiceSimilarities`

### DIFF
--- a/search/tests/helpers.py
+++ b/search/tests/helpers.py
@@ -1,0 +1,24 @@
+from common.testhelpers.random_test_values import a_float
+
+
+def create_scores(test_range):
+    scores = [[a_float() for i in range(test_range)] for j in range(test_range)]
+    
+    if check_if_scores_has_duplicate(scores):
+        create_scores(test_range)
+    return scores
+
+
+def check_if_scores_has_duplicate(scores):
+    float_dictionary = {}
+    has_duplicate = None
+    
+    for i in scores:
+        for j in i:
+            if j in float_dictionary:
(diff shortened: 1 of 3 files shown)


search/tests/helpers.py
```diff
@@ -0,0 +1,19 @@
...
+    return scores
+
+
+def is_there_a_duplicate_score_in_square_matrix(scores):
+    float_dictionary = {}
+
+    for i in scores:
+        for j in i:
+            if j in float_dictionary:
+                True
```
rasmus-storjohann-PG: You never add values to the dict, and you need `return`

search/tests/helpers.py
```diff
@@ -0,0 +1,19 @@
+from common.testhelpers.random_test_values import a_float
+
+
+def create_square_matrix_of_unique_floats(number):
+    scores = [[a_float() for i in range(number)] for j in range(number)]
+    
+    while is_there_a_duplicate_score_in_square_matrix(scores):
+        create_square_matrix_of_unique_floats(number)
+    return scores
```
rasmus-storjohann-PG: Or you use `while true` to avoid the problem of the first test and remove recursion.
```py
while true:
      scores = [[a_float() for i in range(number)] for j in range(number)]
      if not is_there_a_duplicate_score_in_square_matrix(scores) # rename function to not have the not there
             return scores
```

Issue #390, opened 2019-07-05
Newcomers guide content: links without http or https are not clickable 


tomy-pg:
Some of the data in the Newcomers Guide links in KO locale are not clickable. 

![Screenshot_20190708-172259_Arrival Advisor](https://user-images.githubusercontent.com/47794846/60850891-132efb80-a1a6-11e9-99f0-bd52e5017db5.jpg)
![Screenshot_20190708-172316_Arrival Advisor](https://user-images.githubusercontent.com/47794846/60850892-132efb80-a1a6-11e9-8777-b302490cfe52.jpg)

AC: 

- [ ] prefix links beginning with www. with https:// 

QA:

- [ ] Go to  the topics _Have your qualifications and credentials assessed_ and _Applying for a driver's licence_ and ensure that the links in the description are clickable

ian-lin-PG:
The reason why the korean language links aren't clickable is because all of the markdown modifiers have been removed during processing. As a result the entire korean translation is basically just pure text.

A related issue in traditional Chinese is that none of the websites have http:// in them which prevents them from converting correctly to links.

Lastly it seems like all the #Headings in Driver lincense page are all gone.

rasmus-storjohann-PG:
Coming up with a regular expressions that detects all URLs is difficult. This is why, when we imported the Newcomers' Guide content, the people doing the importing were instructed to insert `https://` in front of URLs. I guess that instruction wasn't followed always. Using the very rough "URLs start with www" rule will catch many of those slip ups, so it's and OK solution. However, I don't want us to keep going down that route to make that regex increasingly complex and "smart" to catch all URLs. So I would prefer that any URLs that don't have `https://` and don't start with "www" be fixed manually in the content/ repo to have the `https://` prefix.

ian-lin-PG:
In this scenario it would be easy to fix the error manually doing regex replace. Unfortunately, in issue #732 the traditional Chinese translation is much worse. I observed url not terminated properly with a space at the end, thereby absorbing the rest of the sentence as part of the URL. I also seen url that is not prefixed with www. or https:// making it much harder to catch. And in one case the url is split in the middle.

We should probably decide the process on fixing the translation errors. In cases like this where a simple replacement can be done, we could commit the fix ourselves. Or in cases like #732 have translator reevaluate the links.

ian-lin-PG:
We have chose to resolve this issue manually. Related to issue#739

tomy-pg:
[...]

PR #394, opened 2019-07-12
Issue #390 Insert https:// to url starting with www

ian-lin-PG:


newcomers_guide/clean_data.py
```diff
@@ -101,13 +101,19 @@ def unprotect_newlines(text):
...
 
 
 def clean_up_email_links(text):
     return re.sub(r'([a-zA-Z]+@[^@\s]+\.[^@\s]+[a-zA-Z])', r'Email: [\1](mailto:\1)', text)
 
 
+def add_https_to_somelinks(text):
+    if re.search(r'([^/])([w]{3})', text):
+        text = re.sub(r'([^/])([w]{3})', r'\1https://\2', text)
+    return text
```
rasmus-storjohann-PG: I know this is WIP but I'll give some pointers anyway. 
* `[w]{3}` is a very complicated way of saying `www`. 
* DRY also means make a local variable for the regex and use it in both the test and the substitution.
* Test driven development will shine here, as you add more test cases you will refine this function to properly handle these links at the start and end of strings, etc. Try to come up with the smallest possible set of test cases that self-evidently covers all the possible situations.

ian-lin-PG: I refactored the regex expression out and gave it a more explicit name. I also made the regex to check for the whole "https://" rather than just the "/", again for clarity. I renamed the tests, but I think that's the area where I might need more feedbacks. Will consult @tomy-pg .

rasmus-storjohann-PG:
Closing out this PR, after review of a wider range of issues in the Newcomer's Guide content, we decided on more manual approach using a review tool written last year.

Issue #419, opened 2019-08-07
Do not import addresses and phone numbers from BC211 marked confidential


rasmus-storjohann-PG:
BC211 data set contains physical address, mailing address and phone number records marked with `Confidential="true"`. These records should not be imported.

### Acceptance Criteria 
- [x]  Physical addresses marked with `Confidential="true"` are not imported
- [x]  Mailing addresses marked with `Confidential="true"` are not imported
- [x]  Phone numbers marked with `Confidential="true"` are not imported

### Steps for QA: 
1. Search a recent BC211 data set for records marked confidential.
1. Import the BC211 data set.
1. Confirm that the confidential records are not present in the database after import.

rasmus-storjohann-PG:
QA:
Query with `select * from locations_locationaddress, addresses_address where locations_locationaddress.address_id=addresses_address.id and location_id='...';`

Finding some issues in the BC211 data set:
* 55823278, 9496773, 9488881 has the same address marked as both confidential and not

PR #422, opened 2019-08-08
Issue #419 do not import confidential records

ian-lin-PG:


bc211/tests/test_parser.py
```diff
@@ -488,6 +488,20 @@ def test_does_not_parse_phone_with_empty_phone_number(self):
...
+            '''
+            <Site>
+                <Phone TollFree="false" Confidential="true">
+                    <PhoneNumber>911</PhoneNumber>
+                    <Type>Phone1</Type>
+                </Phone>
+            </Site>'''
+        )
+        phone_numbers = parser.parse_site_phone_number_list(root, site_id)
+        self.assertEqual(len(phone_numbers), 0)
```
rasmus-storjohann-PG: Here I would add two sites to the XML, one with true and one with false, and assert that only one of them was parsed.

bc211/parser.py
```diff
@@ -300,8 +307,8 @@ def clean_phone_number(phone_number):
     else:
         return phone_number
 
-def phone_has_number_and_type(phone):
-    return parse_optional_field(phone, 'PhoneNumber') and parse_optional_field(phone, 'Type')
+def phone_has_number_and_type_and_not_confidential(phone):
```
rasmus-storjohann-PG: The name strongly suggests that the SRP is broken. Sometimes renaming can fix that, `is_valid_phonenumber` is fine by me. Note though that it is only exceptionally that a rename fixes a SRP violation.

bc211/parser.py
```diff
@@ -56,6 +56,13 @@ def parse_optional_field(parent, field):
     return remove_double_escaped_html_markup(value.text)
 
 
+def parse_attribute(parent, attribute):
+    value = parent.get(attribute)
+    if value is None:
+        return None
+    return remove_double_escaped_html_markup(value)
```
rasmus-storjohann-PG: Functions should be located below and as close as possible to the function that calls it.

bc211/parser.py
```diff
@@ -300,8 +307,8 @@ def clean_phone_number(phone_number):
     else:
         return phone_number
 
-def phone_has_number_and_type(phone):
-    return parse_optional_field(phone, 'PhoneNumber') and parse_optional_field(phone, 'Type')
+def phone_has_number_and_type_and_not_confidential(phone):
+    return parse_optional_field(phone, 'PhoneNumber') and parse_optional_field(phone, 'Type') and parse_attribute(phone, 'Confidential') == 'false'
```
rasmus-storjohann-PG: Long line. Also, there are some possible variants to the test. x == 'false' vs x != 'true' vs case-insensitive versions of each of those. Which do you think is the best and why?

bc211/tests/test_parser.py
```diff
@@ -398,6 +398,34 @@ def test_sets_country_to_none_when_missing(self):
         site_id = a_string()
         self.assertIsNone(parser.parse_address(root.find('MailingAddress'), site_id, address_type_id))
 
+    def test_not_parse_confidential_mailing_address(self):
```
rasmus-storjohann-PG: Test names should read as grammatically correct sentences. Sometimes it's OK to leave out the "test_" prefix and just make sure the remainder of the test name is grammatically correct.

bc211/parser.py
ian-lin-PG: The function is being called in 2 different places, in this case it's far from either one of them. I decided to to keep it close to parse_optional_field() cause that's where both addresses and phone number drill down to in the parsing at some point.

bc211/parser.py
ian-lin-PG: Airing on the more conservative side I should have x== 'false'. In the event someone was to misspell 'true' to 'tru', the record will not be kept. if I had gone with x!='true' then it's more liberal in accepting any typos as valid records.

I will add the `.lower()` to the parsed attribute before comparing it to 'false', there by making the check case insensitive.

bc211/tests/test_parser.py
```diff
@@ -516,7 +516,7 @@ def test_does_not_parse_phone_with_empty_phone_number(self):
         phone_numbers = parser.parse_site_phone_number_list(root, site_id)
         self.assertEqual(len(phone_numbers), 0)
 
-    def test_does_not_parse_phone_with_confidential_set_to_true(self):
+    def test_does_not_parse_phone_labeled_confidential(self):
```
rasmus-storjohann-PG: phone_number

bc211/parser.py
```diff
@@ -221,14 +221,23 @@ def parse_postal_address(site_xml, site_id):
...
+    
 def parse_address(address, site_id, address_type_id):
     city = parse_city(address)
     if not city:
         return None
     country = parse_country(address)
     if not country:
         return None
+    confidential = parse_attribute(address, 'Confidential')
+    if confidential == 'true':
```
rasmus-storjohann-PG: `!= false`

PR #427, opened 2019-08-20
Feature/issue 423 search provider

rasmus-storjohann-PG:


rasmus-storjohann-PG:
The json file should not be checked in.

PR #433, opened 2019-08-30
Feature/issue 420 general import script for manual recommendations

rasmus-storjohann-PG:


search/management/commands/manage_manual_recommendations.py
```diff
@@ -50,3 +52,19 @@ def build_change_records(topic_id, service_id_index, exclude_index, csv_data):
             'exclude' : line[exclude_index],
         })
     return result
+
+def save_changes_to_database(change_records):
+    for record in change_records:
+        if record['exclude'] != 'Exclude':
```
ian-lin-PG: Coming back to here if it isn't addressed in later commits.

search/read_similarities.py
```diff
@@ -1,15 +1,5 @@
 import csv
 
-
-def read_manual_similarities(csv_path):
```
tomy-pg: `search/management/commands/set_manual_similarity_scores.py` is affected by this change. 
We need to add the import of this function as: 
`from search.read_manual_similarities import read_manual_similarities`

search/management/commands/manage_manual_recommendations.py
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return list(map(make_record, rows))
+
+def remove_row_count_line(rows):
+    invalid_line_pattern = "\\(\\d+ rows\\)"
+    is_valid = lambda row: not re.match(invalid_line_pattern, str(row[0]))
+    return filter(is_valid, rows)
+
+def filter_included_records(change_records):
+    return list(filter(lambda record: record['exclude'] != 'Exclude', change_records))
```
tomy-pg: If we are indicating that a record should be _excluded_, we should be removing that record from the database as opposed to leaving the record in the database if it already exists.

search/management/commands/manage_manual_recommendations.py
```diff
@@ -0,0 +1,107 @@
...
+from search.models import TaskServiceSimilarityScore
+
+class Command(BaseCommand):
+    help = ('Given a path to a directory, this script reads all CSV files from that '
+            'directory as manual recommendations of services for topics. Format: The '
+            'filenames must match corresponding topic ids, the content of the files are '
+            'CSV files with a column headed "service_id" and another column headed '
+            '"Include/Exclude". Values from these columns are used to create recommended '
+            'service records for the given topic. All other columns are ignored. All rows '
+            'where the "Include/Exclude" column contains "Exclude" are also ignored. All '
```
tomy-pg: This will need to be changed to something like: 

> All rows where the Include/Exclude column contains "Exclude" will be removed from the database if it currently exists. Otherwise, this row will be ignored.

PR #458, opened 2019-10-01
feature/IssueQa#14 Create data models for QA tool

ian-lin-PG:


qa_tool/models.py
```diff
@@ -1,3 +1,10 @@
 from django.db import models
+from common.models import (RequiredURLField,
+                           OptionalTextField, RequiredCharField)
 
-# Create your models here.
+
+class Algorithm(models.Model):
+    id = RequiredCharField(max_length=200, primary_key=True)
```
rasmus-storjohann-PG: Django will add this field by default if you don't define it. Unless you have particular constraints on the field (e.g. because it is being populated from some existing data source like BC211), it's nicer to leave it out and get the default field from Django.

qa_tool/tests/tests.py
```diff
@@ -0,0 +1,39 @@
...
+from common.testhelpers.database import validate_save_and_reload
+from qa_tool.tests.helpers import AlgorithmBuilder
+
+
+class TestAddressModel(TestCase):
+
+    def test_has_name(self):
+        alg = AlgorithmBuilder().build()
+        alg_from_db = validate_save_and_reload(alg)
+        self.assertEqual(alg_from_db.name, alg.name)
```
rasmus-storjohann-PG: Avoid abbreviations in your code. See Uncle Bob.

qa_tool/tests/tests.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_has_notes(self):
+        alg = AlgorithmBuilder().build()
+        alg_from_db = validate_save_and_reload(alg)
+        self.assertEqual(alg_from_db.notes, alg.notes)
+
+    def test_country_field_can_be_empty(self):
+        alg = AlgorithmBuilder().with_notes('').build()
+        alg_from_db = validate_save_and_reload(alg)
+        self.assertEqual(alg_from_db.notes, '')
```
rasmus-storjohann-PG: Nice

qa_tool/models.py
ian-lin-PG: Now I am wondering if it's worth deleting and remigrating just for this.

qa_tool/tests/tests.py
ian-lin-PG: Expanded all alg references to algorithm.

qa_tool/tests/tests.py
ian-lin-PG: Renamed the test lol

qa_tool/models.py
ian-lin-PG: Deleted the id field

qa_tool/models.py
```diff
@@ -0,0 +1,25 @@
...
+    point = models.PointField(blank=True, null=True)
+
+
+class Score(models.Model):
+    value = models.IntegerField()
+    time_stamp = models.DateTimeField()
+    algorithm = models.ForeignKey(Algorithm, on_delete=models.PROTECT)
+    search_location = models.ForeignKey(SearchLocation, on_delete=models.PROTECT)
+    user = models.ForeignKey(User, on_delete=models.PROTECT)
+    service = models.ForeignKey(Service, on_delete=models.PROTECT)
```
rasmus-storjohann-PG: Not `service`, `service_at_location`.

qa_tool/tests/tests.py
```diff
@@ -71,13 +69,12 @@ def test_can_create_row(self):
         time_stamp = timezone.now()
         algorithm = AlgorithmBuilder().create()
         search_location = SearchLocationBuilder().create()
-        organization = OrganizationBuilder().create()
-        service = ServiceBuilder(organization).create()
+        service_at_location = ServiceAtLocationBuilder().create_many()[0]
```
rasmus-storjohann-PG: You don't need to change this in this PR, but for the future, please make changes to the builders to meet your needs. This is how the builders are created and evolve to make future tests easier to write. In this case I think you would have used a `create()` function if it existed. Well, make it, the future you will thank you. @tomy-pg , please note this.

PR #468, opened 2019-10-18
Feature/qa tool issue#15 Set up QA tool REST endpoints with authentication

ian-lin-PG:


qa_tool/tests/helpers.py
```diff
@@ -59,3 +61,32 @@ def create(self):
         result = self.build()
         result.save()
         return result
+
+
+class RelevancyScoreBuilder():
+    def __init__(self, User):
+        self.user = User
```
rasmus-storjohann-PG: Variables should start with lower case

qa_tool/tests/helpers.py
```diff
@@ -59,3 +61,32 @@ def create(self):
...
+        self.value = an_integer()
+        self.time_stamp = timezone.now()
+        self.algorithm = AlgorithmBuilder().create()
+        self.search_location = SearchLocationBuilder().create()
+        self.service_at_location = ServiceAtLocationBuilder().create()
```
rasmus-storjohann-PG: If and when you need to write tests that rely on particular properties on these three related models, you may need to pass these in as you currently do with the user. But for now, this is fine. Remember that test helper code like this is always subject to gradual improvements and expansion to meet the needs of test code.

qa_tool/tests/test_post_new_score.py
```diff
@@ -0,0 +1,53 @@
...
+        self.data = {
+            'id': 1,
+            'user': self.user,
+            'value': 2,
+            'time_stamp': 'http://www.example.org',
+            'algorithm_id': 3,
+            'search_location_id': 4,
+            'service_at_location_id': 5,
+            'user_id': 6
+        }
```
rasmus-storjohann-PG: `self.data` is not used.

qa_tool/tests/test_post_new_score.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_can_get_entities(self):
+        RelevancyScoreBuilder(self.user).with_value(1).create()
+        RelevancyScoreBuilder(self.user).with_value(3).create()
+        url = '/v1/relevancyscores/'
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.json()), 2)
```
rasmus-storjohann-PG: Same as before, `1`, `3` are not relevant. Do you want to assert on the result to make sure it makes it through?

qa_tool/tests/test_post_new_score.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_can_get_one_entity(self):
+        score = RelevancyScoreBuilder(self.user).with_value(3).build()
+        score.save()
```
rasmus-storjohann-PG: Use `create()`

qa_tool/tests/test_search_location.py
```diff
@@ -0,0 +1,47 @@
...
+    def setUp(self):
+        self.data = {
+            'id': 'The id',
+            'name': 'The name',
+            'point': '1 ,2 '
+        }
+
+    def test_can_get_entities(self):
+        SearchLocationBuilder().with_name("wonderland").create()
+        SearchLocationBuilder().with_name("playland").create()
```
rasmus-storjohann-PG: Drop hard coded strings

qa_tool/tests/test_search_location.py
```diff
@@ -0,0 +1,47 @@
...
+        url = '/v1/searchlocations/'
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.json()), 2)
+
+    def test_can_get_one_entity(self):
+        searchlocation = SearchLocationBuilder().with_name('HiVE').build()
+        searchlocation.save()
+        url = '/v1/searchlocations/{0}/'.format(searchlocation.pk)
```
rasmus-storjohann-PG: Again the name string is not making the test any clearer. You could use this pattern:
```
searchlocation = SearchLocationBuilder().with_id(id).create()
url = '/v1/searchlocations/{0}/'.format(id)
...
```

search/tests/test_api.py
```diff
@@ -5,6 +5,47 @@
...
+        }
+
+    def test_can_get_entities(self):
+        create_topic('1')
+        create_topic('2')
+        url = '/v1/topics/'
+        response = self.client.get(url)
+        print(response)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.json()), 2)
```
rasmus-storjohann-PG: Same as before

qa_tool/serializers.py
```diff
@@ -16,6 +16,9 @@ class Meta:
 
 
 class SearchLocationSerializer(serializers.HyperlinkedModelSerializer):
+    latitude = serializers.ReadOnlyField(source='point.y')
+    longitude = serializers.ReadOnlyField(source='point.x')
```
rasmus-storjohann-PG: Make sure you don't get them backwards!

config/urls.py
```diff
@@ -40,7 +40,9 @@ def build_router():
     router.register(r'services_at_location', ServiceAtLocationViewSet, base_name='services_at_location')
     router.register(r'topics/(?P<topic_id>[\w-]+)/related_topics', RelatedTopicsViewSet, base_name='topics')
     router.register(r'topics/(?P<topic_id>[\w-]+)/related_services', RelatedServicesViewSet, base_name='topics')
-    router.register(r'relevancyscores', RelevantScoreViewSet, base_name='relevancyscores')
+    router.register(r'algorithms', AlgorithmViewSet, base_name='algorithms')
+    router.register(r'algorithms/(?P<algorithm_id>[\w-]+)/relevancyscores',
+                    RelevancyScoreViewSet, base_name='relevancyscores')
```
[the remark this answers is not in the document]
ian-lin-PG: renamed everything to relevancyscore for now.

qa_tool/tests/test_algorithm.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class AlgorithmsTests(rest_test.APITestCase):
+    def setUp(self):
+        self.data = {
+            'id': 'The id',
+            'name': 'The name',
+            'url': 'http://www.example.org',
+            'notes': 'The notes'
+        }
```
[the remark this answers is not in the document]
ian-lin-PG: data is being used in post, but it's not meaningful because the post action is not being permitted with the current ReadOnlyModelViewSet.

qa_tool/tests/test_algorithm.py
```diff
@@ -0,0 +1,48 @@
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.json()), 2)
+
+    def test_can_get_one_entity(self):
+        algorithm = AlgorithmBuilder().with_notes('Algorithm notes').build()
+        algorithm.save()
+        url = '/v1/algorithms/{0}/'.format(algorithm.pk)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.json()['notes'], 'Algorithm notes')
```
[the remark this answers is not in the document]
ian-lin-PG: replaced 'Algorithm notes' with a_string(), however, I found most APItests uses pk for formatting. I ran the tests, and the test passes either with id or pk.

qa_tool/tests/test_post_new_score.py
ian-lin-PG: I was under the impression that this way, the 'score' variable gets access to 'pk' field. Replacing .build().save() with create() does not break the tests.

qa_tool/tests/helpers.py
ian-lin-PG: It must be sample codes I copied, changed and all the tests still passed.

qa_tool/tests/test_post_new_score.py
ian-lin-PG: Still have to figure out whether the data should be meaningful for post()

qa_tool/tests/test_post_new_score.py
ian-lin-PG: replaced the 1,3 with an_integer()

qa_tool/tests/test_search_location.py
ian-lin-PG: I got a weird comment in the builder() when I tried to add with_id, something to explore more.

qa_tool/serializers.py
ian-lin-PG: Double checked!

config/settings/base.py
```diff
@@ -61,6 +62,9 @@
     'PAGE_SIZE': 30,
     'ORDERING_PARAM': 'sort_by',
     'DEFAULT_FILTER_BACKENDS': ['common.filters.SearchFilter', 'common.filters.MultiFieldOrderingFilter', ],
+    'DEFAULT_AUTHENTIFICATION_CLASSES': (
```
tomy-pg: Re-word _authentification_ to authentication.

rasmus-storjohann-PG:
Update new admin.py file to register models, see other similar files.

qa_tool/tests/test_algorithm.py
```diff
@@ -0,0 +1,48 @@
...
+            'notes': 'The notes'
+        }
+
+    def test_can_get_entities(self):
+        AlgorithmBuilder().with_name("staging").create()
+        AlgorithmBuilder().with_name("production").create()
+        url = '/v1/algorithms/'
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.json()), 2)
```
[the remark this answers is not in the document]
rasmus-storjohann-PG: Remove test.

qa_tool/tests/test_relevancy_score.py
```diff
@@ -0,0 +1,86 @@
...
+        score = RelevancyScoreBuilder(self.user).with_value(score_value).create()
+        url = '/v1/relevancyscores/{0}/'.format(score.pk)
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.json()['value'], score_value)
+
+    def test_can_post(self):
+        url = '/v1/relevancyscores/'
+        response = self.client.post(url, self.data)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```
rasmus-storjohann-PG: Given that there will be some logic in the post call related to user name and time stamp, you will want one test here for each field, so check that each field makes it through either from the DTO, the token or Time.now().

rasmus-storjohann-PG: Also I'd like to see tests for unauthenticated POST, PUT and DELETE calls, to see that they fail and leave the data unchanged.

config/urls.py
```diff
@@ -50,6 +58,8 @@ def build_router():
     url(r'^about/$', TemplateView.as_view(template_name='pages/about.html'), name='about'),
     url(r'^version/$', VersionView.as_view(), name='version'),
     url(r'^bc211version/$', Bc211VersionView.as_view(), name='bc211_version'),
+    url('api-token-auth/', views.obtain_auth_token, name='api-token-auth'),
+    url(r'^authenticate/', CustomObtainAuthToken.as_view()),
```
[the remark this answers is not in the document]
ian-lin-PG: one of them will remain for the login to hand back the tokens.

qa_tool/tests/test_relevancy_score.py
rasmus-storjohann-PG: Also, change the URL from `/v1/....` to be `/qa/v1/...`

qa_tool/tests/test_relevancy_score.py
```diff
@@ -0,0 +1,104 @@
...
+        response = self.client.delete(url)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_can_delete_just_by_pk(self):
+        score = RelevancyScoreBuilder(self.user).create()
+        RelevancyScoreBuilder(self.user).create()
+        url = '/qa/v1/relevancyscores/{0}/'.format(score.pk)
+        self.client.force_login(user=self.user)
+        response = self.client.delete(url)
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
```
rasmus-storjohann-PG: Not sure why this test is different from `test_can_delete()` above. Also not sure why it needs to use two `RelevancyScoreBuilder()`

qa_tool/tests/test_relevancy_score.py
```diff
@@ -0,0 +1,104 @@
...
+            'search_location': SearchLocationBuilder().create().id,
+            'service_at_location': ServiceAtLocationBuilder().create().id,
+            'topic': create_topic('test').id,
+        }
+
+    def test_can_post(self):
+        algorithm = AlgorithmBuilder().create()
+        url = '/qa/v1/algorithms/{0}/relevancyscores/'.format(algorithm.pk)
+        response = self.client.post(url, self.data)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```
rasmus-storjohann-PG: Here I would expect a get call or a database query to verify that the thing you asked to create really exists

qa_tool/viewsets.py
```diff
@@ -0,0 +1,41 @@
...
+        rawdata = request.data.copy()
+        if 'algorithm' not in rawdata:
+            rawdata['algorithm'] = algorithm_id
+        rawdata['time_stamp'] = timezone.now()
+        rawdata['user'] = request.user.id
+        serializer = serializers.RelevancyScoreSerializer(data=rawdata)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
rasmus-storjohann-PG: - [ ] Algo id can be in the url (success), or in the body (success), or in both (error 400), or in neither (error 400).
- [ ] Test that the time stamp is set correctly
- [ ] Test that the user id is set
- [ ] Test failure of serializer, pass in a string where there should be a number

qa_tool/viewsets.py
```diff
@@ -0,0 +1,41 @@
...
+    queryset = models.Algorithm.objects.all()
+    serializer_class = serializers.AlgorithmSerializer
+
+
+class SearchLocationViewSet(viewsets.ReadOnlyModelViewSet):
+    queryset = models.SearchLocation.objects.all()
+    serializer_class = serializers.SearchLocationSerializer
+
+
+class CustomObtainAuthToken(ObtainAuthToken):
```
rasmus-storjohann-PG: Delete

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: I wanted to show that it was deleting 1 entry vs delete all entries. But the response of a delete doesn't return what's left in the database, will delete the second test.

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: Verifying the algorithm id that was used in the url is the same.

qa_tool/viewsets.py
ian-lin-PG: I have the time_stamp test to be that the time from the database is less than the current time. `assertAlmostEqual` does not work on datetime objects because they do not have an inherent `round()` method.

qa_tool/tests/test_relevancy_score.py
```diff
@@ -35,20 +36,75 @@ def test_can_get_when_not_authenticated(self):
...
+        self.assertEqual(response.json()['user'], self.user.id)
+
+    def test_response_has_similar_time_stamp(self):
+        url = '/qa/v1/relevancyscores/'
+        self.client.force_login(user=self.user)
+        response = self.client.post(url, self.data)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        recreated_date = datetime.strptime(response.json()['time_stamp'], '%Y-%m-%dT%H:%M:%S.%fZ')
+        self.assertLessEqual(recreated_date, datetime.now())
```
rasmus-storjohann-PG: This assert will pass even if the `recreated_date` is way less than now. I would like to see that `recreated_date` is between `now` and `now` less 1 second.

qa_tool/viewsets.py
```diff
@@ -13,8 +11,12 @@ class RelevancyScoreViewSet(viewsets.ModelViewSet):
 
     def create(self, request, algorithm_id=None, *args, **kwargs):
         rawdata = request.data.copy()
-        if 'algorithm' not in rawdata:
+        if 'algorithm' not in rawdata and algorithm_id is not None:
             rawdata['algorithm'] = algorithm_id
+        elif 'algorithm' not in rawdata and algorithm_id is None:
+            return Response('Algorithm id missing in payload', status=status.HTTP_400_BAD_REQUEST)
+        elif 'algorithm' in rawdata and algorithm_id is not None:
+            return Response('Duplicate sources of algorithm id when there should only be 1', status=status.HTTP_400_BAD_REQUEST)
```
rasmus-storjohann-PG: SRP: The figuring out what the valid algorithm is "one thing" and should be pulled out into a separate function.

qa_tool/tests/test_relevancy_score.py
```diff
@@ -22,122 +26,123 @@ def test_can_get_one_entity(self):
...
+        response = self.APIClient.post(url, self.data)
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
         self.assertEqual(response.json()['search_location'], self.data['search_location'])
 
     def test_response_has_correct_service_at_location(self):
         url = '/qa/v1/relevancyscores/'
-        self.client.force_login(user=self.user)
-        response = self.client.post(url, self.data)
+        self.APIClient.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
+        response = self.APIClient.post(url, self.data, HTTP_AUTHORIZATION='Toke {}'.format(self.token))
```
rasmus-storjohann-PG: Spelling: Toke. More examples below. I'm also not sure what this line of code does.

qa_tool/tests/test_relevancy_score.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def test_can_get_entities(self):
+        RelevancyScoreBuilder(self.user).create()
+        RelevancyScoreBuilder(self.user).create()
+        url = '/qa/v1/relevancyscores/'
+        response = self.APIClient.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.json()), 2)
+
+    def test_can_get_when_not_authenticated(self):
```
rasmus-storjohann-PG: All the GET tests are not authenticated, so why is this test needed?

qa_tool/tests/test_relevancy_score.py
```diff
@@ -0,0 +1,255 @@
...
+        self.APIClient.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
+        response = self.APIClient.put(url, self.data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_cannot_put_when_not_authenticated(self):
+        score = RelevancyScoreBuilder(self.user).create()
+        url = '/qa/v1/relevancyscores/{0}/'.format(score.pk)
+        self.data['value'] = an_integer()
+        response = self.APIClient.put(url, self.data)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
```
rasmus-storjohann-PG: PUT and POST should also return the resulting value, using the same format at GET returns.

qa_tool/tests/test_relevancy_score.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def test_can_post_with_algorithm_in_body_not_in_url_short_url(self):
+        url = '/qa/v1/relevancyscores/'
+        self.APIClient.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
+        response = self.APIClient.post(url, data=self.data_with_algorithm_in_body)
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.json()['value'], self.score_value)
+
+    def test_cannot_post_with_algorithm_in_body_not_in_url_long_url(self):
+        url = '/qa/v1/algorithms/not_algorithm_id/relevancyscores/'
```
rasmus-storjohann-PG: Test name is misleading, the algorithm id *is* in the URL, it's just an invalid one. This end point should return an error even with a valid id in the url and the body.

qa_tool/tests/test_relevancy_score.py
```diff
@@ -0,0 +1,255 @@
...
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_cannot_post_without_algorithm_in_body_or_url_short_url(self):
+        url = '/qa/v1/relevancyscores/'
+        self.APIClient.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
+        response = self.APIClient.post(url, data=self.data_without_algorithm_in_body)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_cannot_post_without_algorithm_in_body_or_url_long_url(self):
+        url = '/qa/v1/algorithms/not_algorithm_id/relevancyscores/'
```
rasmus-storjohann-PG: This call should succeed, there is no reason why "not_algorithm_id" isn't a valid id. If the problem is that is needs to be a number, then this test is testing validation logic, not the conditions on the id being present in one or the other part of the request. Makes sense?

qa_tool/tests/test_relevancy_score.py
```diff
@@ -0,0 +1,255 @@
...
+        self.APIClient.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
+        response = self.APIClient.post(url, data=self.data_with_algorithm_in_body)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_cannot_post_with_different_algorithm_in_body_and_url_short_url(self):
+        algorithm_id = an_integer()
+        url = '/qa/v1/relevancyscores/{0}/'.format(algorithm_id)
+        self.APIClient.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
+        response = self.APIClient.post(url, data=self.data_with_algorithm_in_body)
+        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
```
rasmus-storjohann-PG: There are a couple of extraneous tests here, with the same vs. different ids in URL and body ... the rule is simply related to whether the ids are there, don't care of they're the same or different.

qa_tool/tests/test_relevancy_score.py
rasmus-storjohann-PG: ???

qa_tool/tests/test_relevancy_score.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def test_cannot_post_with_different_algorithm_in_body_and_url_long_url(self):
+        algorithm_id = an_integer()
+        url = '/qa/v1/algorithms/{0}/relevancyscores/'.format(algorithm_id)
+        self.APIClient.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
+        response = self.APIClient.post(url, data=self.data_with_algorithm_in_body)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_can_post_with_algorithm_in_url_not_in_body_long_url(self):
```
rasmus-storjohann-PG: For improved readability, more the two `test_can_post_...` together up above all the `test_cannot_post_...` tests, so that we understand what specifically can be done before we get into all the things that can't

qa_tool/viewsets.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def update(self, request, pk, *args, **kwargs):
+        score = models.RelevancyScore.objects.get(pk=pk)
+        request_data = request.data.copy()
+        request_data = self.attach_missing_info(request_data, request.user.id)
+        serializer = serializers.RelevancyScoreSerializer(score, data=request_data)
+        if not serializer.is_valid():
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+        serializer.save()
+        return Response(serializer.data)
```
rasmus-storjohann-PG: There is an implicit merge that happens in a PUT, where data from the request are combined with the data already in the database. There are no tests to show how this happens. Can I update any field using a PUT? Is the timestamp updated? Is the user updated? Can I change any of the foreign keys or just they value?

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: oh I added verification on the algorithm id after the post was complete.

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: This is to differentiate from `ReadOnlyOrIsAuthenticated` from `IsAuthenticated`. The latter case is more restrictive and would not get calls if the request is not authenticated. Perhaps I need to rename the test to reflect this distinction.

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: I recorded the new value for put and compared with the response. I left the HTTP_401 as is because it does not return a json body.

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: I deleted the said test and mention explicitly every time if a test case is `unauthenticated`.

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: I have decided to change all `not_algorithm_id` or equivalent to be 0 to be consistent.

ian-lin-PG: I have decided to include the old tests but rename them to be more appropriate

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: Removed the duplicate tests.

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: Shuffled the tests such that `test_can_post...` is followed by `test_response...` and lastly `test_cannot_post`. The rationale is that all success or 200 tests would go before all the failing tests.

qa_tool/viewsets.py
ian-lin-PG: added tests to every single field to verify if they can be updated through `put`

qa_tool/tests/test_relevancy_score.py
```diff
@@ -90,27 +92,95 @@ def setUp(self):
...
+        self.assertLessEqual(recreated_time, datetime.datetime.now())
+        self.assertGreaterEqual(recreated_time + datetime.timedelta(seconds=1), datetime.datetime.now())
+
+    def test_can_put_with_different_credential(self):
+        new_token = Token.objects.create(user=self.new_user)
+        self.APIClient.credentials(HTTP_AUTHORIZATION='Token ' + new_token.key)
+        url = '/qa/v1/relevancyscores/{0}/'.format(self.sample_score.pk)
+        self.data['value'] = an_integer()
+        response = self.APIClient.put(url, self.data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
rasmus-storjohann-PG: add assert to check that the new user is set on the record...

qa_tool/tests/test_relevancy_score.py
```diff
@@ -90,27 +92,95 @@ def setUp(self):
...
+
+    def test_cannot_put_id(self):
+        self.APIClient.credentials(HTTP_AUTHORIZATION='Token ' + self.token.key)
+        new_id = an_integer()
+        self.data['id'] = new_id
+        url = '/qa/v1/relevancyscores/{0}/'.format(self.sample_score.pk)
+        response = self.APIClient.put(url, self.data)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json(), 'id of relevancyscore is immutable')
```
rasmus-storjohann-PG: nice

qa_tool/admin.py
```diff
@@ -1,3 +1,6 @@
 from django.contrib import admin
+from .models import Algorithm, SearchLocation, RelevancyScore
```
rasmus-storjohann-PG: Avoid relative paths in django

qa_tool/tests/test_relevancy_score.py
ian-lin-PG: Also asserted to check the value was actually changed

PR #471, opened 2019-11-09
Feature/issue 441 post end point for push notifications

rasmus-storjohann-PG:


push_notifications/view_sets.py
```diff
@@ -0,0 +1,39 @@
...
+
+@swagger_auto_schema(
+    methods=['PUT'],
+    request_body=PushNotificationTokenSerializer,
+    responses={
+        200: PushNotificationTokenSerializer(many=False),
+        400: 'Bad request',
+    },
+    operation_description='Save a push notification token of form "ExponentPushToken[xyz]"')
+@api_view(['PUT'])
```
ian-lin-PG: Is there a reason for picking functional based views vs using class based views?

rasmus-storjohann-PG: Yes, the class based view is a very compact package solution that gives you the GET, POST, PUT and DELETE in one thing. Here, I just want a PUT, so a functional view is a better solution. Django is really a system for making it really easy to build the bog standard REST end points, but this token end point is not so standard, it's actually quite unusual, so it calls for using a more flexible approach than the class based view.

config/urls.py
```diff
@@ -66,21 +67,20 @@ def build_qa_tool_routes():
...
     url(r'^users/', include('users.urls')),
     url(r'^accounts/', include('allauth.urls')),
 
-    # Your stuff: custom urls includes go here
     url(r'^swagger(?P<format>.json|.yaml)$', SCHEMA_VIEW.without_ui(cache_timeout=None), name='schema-json'),
     url(r'^swagger/$', SCHEMA_VIEW.with_ui('swagger', cache_timeout=None), name='schema-swagger-ui'),
     url(r'^redoc/$', SCHEMA_VIEW.with_ui('redoc', cache_timeout=None), name='schema-redoc'),
 
     url(r'^v1/', include(build_router().urls)),
+    url(r'^v1/push_notifications/tokens/(?P<token>[A-Za-z0-9\_\[\]]+)/', create_or_update_push_notification_token),
```
rasmus-storjohann-PG: @bensinjin this is the offending line, the regex here is too restrictive. Should probably make it just `.*`

bensinjin: Ah yeah, I think that's wise. I might even take it a step further and add "ExponentPushToken" as the prefix here, but that's up to you.

bensinjin: I wonder if 404 is the right response here? I sort of think we should remove this sort of url validation instead have a 400 response with a more useful message.

push_notifications/management/commands/send_push_notification.py
```diff
@@ -0,0 +1,97 @@
...
+    except (ConnectionError, HTTPError):
+        LOGGER.error('Error pushing notification: ConnectionError for %s', token)
+
+    try:
+        response.validate_response()
+
+    except DeviceNotRegisteredError:
+        # According to https://github.com/expo/expo-server-sdk-python this error is
+        # thrown when users no longer want to receive push notifications.
+        LOGGER.warning('DeviceNotRegisteredError, remove this token from our database: %s', token)
```
bensinjin: Should the removal of these tokens be implemented before we merge this work? Have you been able to create this error in your testing?

push_notifications/serializers.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework import serializers
+from push_notifications import models
+
+
+class PushNotificationTokenSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.PushNotificationToken
+        fields = '__all__'
```
bensinjin: Is this class necessary (or even the serializer?) for endpoints we never GET?

Issue #456, opened 2019-09-30
Services with No Lat Long


jparlina:
Based on Sep 16 thread on Slack #pg-product-team, since we use bc211 taxonomy, the recommended services are now sorted by distance instead of similarity score, and are cut off at 25 km max. What do we do with services that have no lat long? 

Example: Under "Community: Volunteering" topic, the service in Whistler appears before the one in Delta (see screenshot). The service is Whistler has no lat long (Service ID53935282) 

![File_001](https://user-images.githubusercontent.com/48343061/65907393-28689000-e379-11e9-91b5-724fc2de815b.png)

### Acceptance Criteria 
- [ ]  

### Steps for QA: 
1.   run the following commands in the pathways-backend repo 
`./utility/prepare_deploy.sh --bc211Path ../bc211/iCarolExport-BC211-AIRSXML-20190613_014810.xml --city_latlongs ../content/city_latlong.csv --newComersGuidePath ../content/NewcomersGuide/ --recommendationsToAddPath ../content/taxonomy/ --outputDir ../`

ian-lin-PG:
Here is the list of services with no latlong https://docs.google.com/spreadsheets/d/1xk3KkoSmiYG_TjSpViMAcvO0_2mDNfDMqnGwp9qUGx0/edit#gid=2112422893
@ArtiBPG please review and close the issue when you think it's complete.

ian-lin-PG:
Out of the 270 services/location without the latlong, the only field that seems to be required is the city. 

I propose that we double check the city within the addresses if the latlong is `NULL` and then use either geocoder or a saved list of city to latlong pairs to add back the latlong at the importer stage. Given that these latlong are inferred instead of known latlongs, appropriate warnings may need to be added to the service listing to indicate as such.

tomy-pg:
QA'd Staging version 1.14.0, Server version 1.3.20:

Under "Community: Volunteering" topic, the service in Whistler still appears before Delta.
I have checked the service and it now does contain a Whistler latitude and longitude. 

While including the cityLatLongs as part of `import_bc211_data`, the script also logged services that did not have latitude and longitude values at the time. 

Will investigate this further.

ian-lin-PG:
[...]

PR #499, opened 2019-12-16
Issue #456 Update prepare_deploy script

ian-lin-PG:


utility/prepare_deploy.sh
```diff
@@ -9,6 +9,10 @@ while (( "$#" )); do
     then
         BC211Path=$2
         shift 2
+    elif [ "$1" == "--city_latlongs" ]
+    then
+        LatlongReplacements=$2
```
tomy-pg: I think the name could be changed to `CityLatLongs`. We are importing a file with latitude and longitude of cities. The implementation detail of replacing can be described within the importer.

utility/prepare_deploy.sh
```diff
@@ -40,6 +44,9 @@ usage() {
     echo "    --bc211Path"
     echo "                The path to the BC211 data set in XML iCarol format."
     echo
+    echo "    --city_latlongs"
```
tomy-pg: Camel case for consistency. `--cityLatLongs`

PR #501, opened 2019-12-20
Feature/issue 500 upgrade libraries

tomy-pg:
This PR is blocked by django-parler. 

As discussed here https://docs.djangoproject.com/en/3.0/releases/3.0/#removed-private-python-2-compatibility-apis:

> django.utils.six and django.utils.encoding.python_2_unicode_compatible are removed in django 3.

Django-parler currently imports six from django.utils and we currently receive this error every time we run the tests. : 

> ImportError: cannot import name 'six' from 'django.utils' (/Users/tomyreyes/Desktop/pathways-backend/.venv/lib/python3.7/site-packages/django/utils/__init__.py)

The maintainers of django-parler have recently dropped their use of six and currently have the changes in their master branch. Will we have to wait until a stable version of django-parler is released with these changes. Discussion on this issue can be found here:
https://github.com/django-parler/django-parler/pull/255

Along with upgrading the libraries, there have been some small changes made in a couple of files because of code becoming deprecated. 

Django's documentation recommends running `python3 -Wa ./manage.py test` for the purposes of removing deprecated code along with becoming familiar with new changes.

### DIFF
--- a/bc211/parser.py
+++ b/bc211/parser.py
@@ -251,7 +251,8 @@ def record_is_confidential(record):
     return True
 
 def parse_address_lines(address):
-    sorted_address_children = sorted(address.getchildren(), key=lambda child: child.tag)
+    address_children = list(address)
+    sorted_address_children = sorted(address_children, key=lambda child: child.tag)
     address_line_tags = [child.tag for child in sorted_address_children if re.match('Line[1-9]', child.tag)]
     if not address_line_tags:
         return None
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: python
-python: 3.6.3
+python: 3.8
```
rasmus-storjohann-PG: 3.7 in the docker file and 3.8 here, is there a good reason for that?

tomy-pg: No I was chasing a lead in fixing the following error: 

```AttributeError: /usr/lib/libgdal.so.1: undefined symbol: OGR_F_GetFieldAsInteger64```

I will change this to make versions more consistent.

PR #505, opened 2020-02-07
Feature/issue 490 update bc211 database

rasmus-storjohann-PG:


bc211/tests/test_updater.py
```diff
@@ -25,6 +27,35 @@ def test_update_existing_location(self):
...
+                            with_phone_numbers([phone_at_location]))
+
+        save_locations([location_builder.build_dto()], {}, ImportCounters())
+
+        new_phone_at_location = dtos.PhoneAtLocation(phone_number_type_id=phone_number_type_id,
+                                                     phone_number=new_phone_number,
+                                                     location_id=location_id)
+        save_locations(
+            [(location_builder.
+              with_phone_numbers([new_phone_at_location]).
```
ian-lin-PG: Would like to see at test with multiple phone numbers for replacement.

bc211/location.py
```diff
@@ -39,7 +39,7 @@ def get_ids_of_locations_to_delete(locations, organization_id):
 
 def delete_locations(location_ids_to_delete):
     for i in location_ids_to_delete:
-        LOGGER.info('delete "%2"', i.id)
+        LOGGER.info('delete "%2"', i)
```
tomy-pg, line 42: I am seeing a logging error during import related to this I believe. Do you mean to use `"%s"` here?

rasmus-storjohann-PG:
There are definitely benefits to using CSV as the source of truth, including being able to keep it in git. Found some reads on this [here](https://www3.dbmaestro.com/blog/how-the-database-is-your-main-agile-development-bottleneck) and [here](https://www3.dbmaestro.com/blog/using-git-for-database-why-and-how).

bc211/location.py
```diff
@@ -21,10 +18,8 @@ def update_locations(locations, organization_id, city_latlong_map, counters):
             save_location(location, None, city_latlong_map, counters)
             counters.count_locations_created()
             LOGGER.info('created "%s" "%s"', location.id, location.name)
```
tomy-pg, line 20: I do not think we need to log when an organization, location or a service is created anymore if we are importing data after flushing the current data in the database.

tomy-pg (changes requested):

PR #515, opened 2020-03-27
Issue pg-irc/pathways-frontend#919 Create a newcomer tag to algolia service data

ian-lin-PG:


utility/algolia/add_phone_numbers_to_services.py
```diff
@@ -53,6 +53,10 @@ def set_taxonomy_terms_on_services(algolia_data, taxonomy_data):
         service_id = service['service_id']
         if service_id in taxonomy_data:
             service['taxonomy_terms'] = taxonomy_data[service_id]
+            if 'immigrantsethnocultural-groups' in taxonomy_data[service_id]:
+                service['for_newcomers'] = 1
+            else:
+                service['for_newcomers'] = 0
```
rasmus-storjohann-PG: I would prefer true/false over 1/0 for this, just because it communicates better what it going on. Just looking at the table and seeing 1 and 0 would make me wonder whether other numbers might be valid values, with true/false I would understand better what's going on.

PR #517, opened 2020-04-09
WIP Issue pg-irc/pathways-frontend#1093 parse last verified date from…

niko-pg:
… Resource Info


rasmus-storjohann-PG:
Really surprised to see so many migration files in this PR, it's a sign that something was not done right. DB migration files are subtle and it's relatively easy to create error states that are really hard to sort out, sometimes this can affect the integrity of the data in the DB. Together with the issue of the large refactor waiting, I'm really not sure if this is a good time for this work to proceed. Sorry. I did not realize the implications when you started working on it.

human_services/services/models.py
```diff
@@ -17,6 +17,7 @@ class Service(ValidateOnSaveMixin, TranslatableModel):
                                             db_table='services_service_taxonomy_terms')
     translations = TranslatedFields(name=models.CharField(max_length=200),
                                     description=models.TextField(blank=True, null=True))
+    last_verified_date = OptionalCharField(max_length=200)
```
rasmus-storjohann-PG, line 20: Let's be more typesafe than that, use a date datatype but allow NULL.

bc211/tests/test_parser.py
```diff
@@ -234,6 +270,12 @@ def test_sets_the_site_id(self):
         self.assertEqual(self.from_minimal_data.site_id,
                          self.site_id_passed_to_parser)
 
+    def test_can_parse_date_last_verified(self):
+        self.assertEqual(self.from_real_data.last_verified_date,
+                         '2016-04-14')
+        self.assertEqual(self.from_minimal_data.last_verified_date,
+                         '2017-03-25')
```
rasmus-storjohann-PG, line 277: What does the parser do on invalid date format?

Issue #527, opened 2020-05-01
Server side: Dynamic content on the home page


rasmus-storjohann-PG:
[This is the client side issue](https://github.com/pg-irc/pathways-frontend/issues/1151).

- [x] We have a new Django app called "content", which contains the following:
- [x] We have a database table for alerts
- [x] Each alert has a heading (localized), a content string (localized) and an id. Localized strings are managed using `TranslatedFields` from `django.parler`, following the pattern in the services table.
- [ ] We can use the Django admin console (web interface) to maintain the content of the table, including localized strings.
- [x] A GET request to /alerts/{locale} returns a list of zero or more alerts in the given locale
- [x] A GET request to /alerts/{locale}/{id} returns the alert with the given id in the given locale
- [x] PUT, POST, DELETE are not supported on these end points
- [ ] Populate the table with data identical to the current COVID-19 card in the app. This is here to demonstrate that the feature works, using a development server install. Populating this table in staging or production is out of scope for this story.

niko-pg:
Hi all, I've started working on this. As of writing, my working model has the following attributes:

```
class Alert(ValidateOnSaveMixin, TranslatableModel):
   id = RequiredCharField(primary_key=True,
                           max_length=200,
                           validators=[validators.validate_slug])

  translations = TranslatedFields(title=models.CharField(max_length=200),
                                    content=models.TextField(blank=True, null=True))
```

Do you have any preliminary thoughts on this @rasmus-storjohann-PG ? I was thinking maybe it should be related to a taxonomy/topic, but in the spirit of "simplify", I didn't consider it for now.

rasmus-storjohann-PG:
@niko-pg note that there's a bit of work left to complete on this story.

rasmus-storjohann-PG:
Sorry, I only saw your questions now, 10 days later. Sorry about that.

All the answers to your questions are in the story, though. If I wanted additional columns I would have spelled that out. Also, the names of the columns were specified but you used different names. So the short answer is really, if it's not specified in the story, then it's out of scope for the PR.

rasmus-storjohann-PG:
I was also working with the PR a bit to adjust naming, so in the end the URLs are a little bit different from the spec. I usually find that finalizing naming needs to be hands on with the code.

Hope this makes sense, we can discuss if you like.

tomy-pg:
Staging version 1.35.0: PASS

PR #528, opened 2020-05-12
Issue #527 add new content app

niko-pg:


config/urls.py
```diff
@@ -43,10 +44,14 @@ def build_router():
     router.register(r'topics', TopicViewSet, basename='topics')
     router.register(r'topics/(?P<topic_id>[\w-]+)/related_topics', RelatedTopicsViewSet, basename='topics')
     router.register(r'topics/(?P<topic_id>[\w-]+)/related_services', RelatedServicesViewSet, basename='topics')
+    router.register(r'alerts/(?P<locale>[\w-]+)', AlertViewSet, basename='alerts')
+    router.register(r'alerts/(?P<locale>[\w-]+)/(?P<alert_id>[\w-]+)', AlertViewSet, basename='alerts')
 
     return router
 
 
+print(build_router().urls)
```
tomy-pg, line 53: did you mean to leave this here?

niko-pg: Removed.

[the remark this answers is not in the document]
niko-pg: Removed

content/tests/test_models.py
```diff
@@ -0,0 +1,76 @@
...
+            alert.full_clean()
+
+    def test_id_cannot_be_empty(self):
+        alert_id = ''
+        alert = AlertBuilder().with_id(alert_id).build()
+        with self.assertRaises(exceptions.ValidationError):
+            alert.full_clean()
+
+    def test_cannot_be_empty(self):
+        alert_id = 'alert id'
```
rasmus-storjohann-PG, line 27: I would normally use `alert_id = a_string()` here. `a_string()` returns a random string, so using it communicates the fact that the content of the string is not relevant to the test, only that the same string shows up in the result.

content/tests/test_models.py
```diff
@@ -0,0 +1,76 @@
...
+        with self.assertRaises(exceptions.ValidationError):
+            alert.full_clean()
+
+    def test_id_cannot_be_empty(self):
+        alert_id = ''
+        alert = AlertBuilder().with_id(alert_id).build()
+        with self.assertRaises(exceptions.ValidationError):
+            alert.full_clean()
+
+    def test_cannot_be_empty(self):
```
rasmus-storjohann-PG, line 26: test what can't be empty?

content/tests/test_models.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.assert_title_in_language_equals(alert_from_db, 'en', 'In English')
+        self.assert_title_in_language_equals(alert_from_db, 'fr', 'En français')
+
+    def assert_title_in_language_equals(self, alert, language, expected_text):
+        alert.set_current_language(language)
+        self.assertEqual(alert.title, expected_text)
+
+    def test_has_desription_field(self):
+        description = 'the description'
```
rasmus-storjohann-PG, line 52: I would normally use `description = a_string()` here. `a_string()` returns a random string, so using it communicates the fact that the content of the string is not relevant to the test, only that the same string shows up in the result.

content/tests/test_models.py
niko-pg: Removed. Seems like a copy paste error

content/tests/test_models.py
niko-pg: Done

common/filters.py
```diff
@@ -177,3 +178,12 @@ def filter_queryset(self, request, queryset, view):
             if queryset.model is Service:
                 queryset = queryset.filter(organization_id=organization_id)
         return queryset
+
+class AlertIdFilter(filters.BaseFilterBackend):
```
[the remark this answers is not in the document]
niko-pg: Oh wow good call. I wonder how the `test_can_get_one_alert` have passed?

content/tests/test_api.py
```diff
@@ -22,11 +22,15 @@ def test_can_get_alerts(self):
...
     def test_can_get_one_alert(self):
+        AlertBuilder().create()
+        AlertBuilder().create()
         alert = AlertBuilder().create()
         url = '/v1/content/alerts/en/{0}/'.format(alert.id)
         response = self.client.get(url)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.json()['id'], alert.id)
+        self.assertEqual(response.json()['heading'], alert.heading)
+        self.assertEqual(response.json()['content'], alert.content)
```
rasmus-storjohann-PG, line 33: TDD can save you from this kind of error. Write the test first, make sure it fails, then write the code to make it pass. A longer test like this could be written in 2-3 of those cycles. The important point is that when you write the code that makes the functionality work, you see a test go from red to green. If you don't see that, you can't be 100% sure that the test is actually testing what you think it's testing.

Issue #465, opened 2019-10-18
Bug in Service Listing Sorting Order by Distance (nearest to furthest)


rasmus-storjohann-PG:
@jparlina commented on [Wed Oct 16 2019](https://github.com/pg-irc/pathways-frontend/issues/877)

Tested on iOS 13.1.3 iPhone 7 plus at ISS of BC Welcome Centre - Surrey (10334 152a St, Suite 301, Surrey, BC V3R 7P8) on both: 
- Production V1.7.0; Server V1.3.12; Jun 2019 bc211 data set
- Staging V1.6.0 (same as 1.7.0 in Production); Server V1.3.13; Jun 2019 bc211 data set

_ISS of BC Welcome Centre - Surrey_ shows up as the 24th service of all 30 services listed under the topic _Contact workers at your local settlement agency_ (see screenshots [here](https://docs.google.com/document/d/1Yrq4tR8lcGxLRm4f6spuV4PVMUsuBV0FjZ9EeETSYZk/edit?usp=sharing))

Checked lat/lon using data:
![Screen Shot 2019-10-16 at 3 20 45 PM](https://user-images.githubusercontent.com/48343061/66965866-13a22280-f030-11e9-955b-c3268cf2497c.png)

Checked lat/lon using WiFi:
![Screen Shot 2019-10-16 at 3 20 56 PM](https://user-images.githubusercontent.com/48343061/66965878-2157a800-f030-11e9-8f56-cd6692807664.png)

Checked lat/long on Google Map:
![Screen Shot 2019-10-16 at 3 21 23 PM](https://user-images.githubusercontent.com/48343061/66965891-2fa5c400-f030-11e9-86e2-40f89de33f5c.png)

### Acceptance Criteria 
- [ ]  When searching services from the physical location of office, the service should be at the top or near the top of the service listing, AND 
- [ ]  services should be sorted by distance - i.e. from nearest to furthest from your location

### Steps for QA: 
1.  Select a topic - e.g. Contact workers from your local settlement agency
2. Click "Find related services near me"
3. Services on the service listing should be sorted by distance - i.e. from nearest to furthest from your location

rasmus-storjohann-PG:
The cause of this problem is that the lat/long data for this location is inaccurate in the BC-211 data set. @ArtiBPG I will put this story into triage for reassessment.

ArtiBPG:
BC211 reported that they corrected this. Check with next export from iCarol.

[...]

PR #534, opened 2020-06-09
Issue #465 Flag out of place services

ian-lin-PG:


utility/add_new_cities_to_csv.py
```diff
@@ -0,0 +1,56 @@
+import csv
```
rasmus-storjohann-PG, line 1: Updates city lat long CSV file with new cities from the BC211 dataset: rename script to `update_csv_with_cities_from_db.py`, add a comment at the top explaining what this script is for and how to call it.

utility/find_out_of_place_services.py
```diff
@@ -0,0 +1,76 @@
+from math import sin, cos, sqrt, atan2, radians
```
rasmus-storjohann-PG, line 1: Rename this script to dump_cities_with_latlong_outside_their_address_city.py

PR #539, opened 2020-06-23
Feature/issue 531 automatic test for search

rasmus-storjohann-PG:


search/tests/test_algolia_search.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_city_in_query(self):
+        if self.is_disabled():
+            print('Algolia tests not run, set environment variable ALGOLIA_SEARCH_API_KEY to enable')
+            return
+        surrey = '49.183333,-122.850000'
+        data = {'query': 'Food', 'page': '1', 'hitsPerPage': '20',
+                'aroundLatLng': surrey, 'aroundPrecision': ''}
+        first_five_results = self.get_search_results(data, 5)
+        service_ids = [result['service_id'] for result in first_five_results]
+        self.assertIn('47982057', service_ids)
```
ian-lin-PG, line 44: `47982057` could use a name like `relevant_service_id` otherwise it looks like a random service_id that could be changed just so the test would pass.

search/tests/test_algolia_search.py
```diff
@@ -0,0 +1,74 @@
...
+    # city as part of the query gives the correct result and that leaving off the city gives the wrong result
+    # Test with a query with lat/long in Surrey close to the surrey/new west border, with a search term that
+    # includes "Surrey", and confirm that we get results that are all in Surrey. See also #919.
+
+    def test_city_in_query(self):
+        if self.is_disabled():
+            print('Algolia tests not run, set environment variable ALGOLIA_SEARCH_API_KEY to enable')
+            return
+        surrey = '49.183333,-122.850000'
+        data = {'query': 'Food', 'page': '1', 'hitsPerPage': '20',
```
ian-lin-PG, line 40: Potentially give `'Food'`a name like `query_term`.

search/tests/test_algolia_search.py
```diff
@@ -0,0 +1,74 @@
...
+    # https://github.com/pg-irc/pathways-frontend/issues/1141 need a test that shows that giving the
+    # city as part of the query gives the correct result and that leaving off the city gives the wrong result
+    # Test with a query with lat/long in Surrey close to the surrey/new west border, with a search term that
+    # includes "Surrey", and confirm that we get results that are all in Surrey. See also #919.
+
+    def test_city_in_query(self):
+        if self.is_disabled():
+            print('Algolia tests not run, set environment variable ALGOLIA_SEARCH_API_KEY to enable')
+            return
+        surrey = '49.183333,-122.850000'
```
ian-lin-PG, line 39: As we write more tests potentially with more city and latlongs, we might need to save the dictionary into `self` for easy reuse. Not necessary for now.

rasmus-storjohann-PG:
We need to be able to check that the search results are sorted correctly, meaning we need to compute the distance between the search location and the location of each search result.

Issue #543, opened 2020-07-28
Simplified and Traditional Chinese Content-alert are using the same server content


tomy-pg:
When content alerts are loaded onto our app, we notice that the strings for Simplified and Traditional Chinese are the same even though they are different on the database. 

The front-end and the back-end use different language codes to identify Simplified and Traditional Chinese. 
**Front-end**:
Simplified Chinese - `zh_CN`
Traditional Chinese - `zh_TW`

**Back-end**
simplified Chinese - `zh-hans`
traditional Chinese - `zh-hant`

Through some debugging, it appears that the front-end pulls the server content from`zh-hans` for both `zh_CN` and `zh_TW`. 

The server should accept both variations of the locale codes described above.

### Acceptance Criteria 
- [x]  When Traditional Chinese is selected on the app, it receives the Traditional Chinese content alerts from the server
- [x] When Simplified Chinese is selected on the app, it receives the Simplified Chinese content alerts from the server

### Steps for QA: 
1.  Populate your local `pathways_local` database with content alerts - be ensure to include rows for `zh-hans` and `zh-hant`.
2. Run your local server and open up swagger
3. Try out the GET request for `/v1/content/alerts/{locale}/` using `zh_TW` as your locale. This is the locale code for Traditional Chinese used by the client.
- [x] Ensure that you receive the content alert for Traditional Chinese - `zh-hant`

ian-lin-PG:
Pass QA at 1.42.0

PR #544, opened 2020-07-28
Issue #543 Convert front-end Chinese language codes to match server locale codes

tomy-pg:


content/tests/test_api.py
```diff
@@ -76,10 +76,26 @@ def test_can_return_result_in_specific_locale(self):
         self.assertEqual(response.json()['heading'], 'Title')
         self.assertEqual(response.json()['content'], 'Description')
 
+    def test_can_return_correct_chinese_translation(self):
+        alert = AlertBuilder().build()
```
rasmus-storjohann-PG, line 80: It would be really nice to be able to do this
```py
alert = AlertBuilder().
              with_heading_in_language('zh-hans', 'Simplified Chinese Title').
              with_content_in_language('zh-hans', 'Simplified Chinese description').
              with_heading_in_language('zh-hant', 'Traditional Chinese Title').
              with_content_in_language('zh-hant', 'Traditional Chinese description').
              build()
```
It does mean the builder gets a bit complex in terms of the data structure it needs to contain...

content/tests/test_api.py
```diff
@@ -76,10 +76,26 @@ def test_can_return_result_in_specific_locale(self):
...
+
+        self.set_heading_in_language(alert, 'zh-hans', 'Simplified Chinese Title')
+        self.set_heading_in_language(alert, 'zh-hant', 'Traditional Chinese Title')
+
+        self.set_content_in_language(alert, 'zh-hans', 'Simplified Chinese description')
+        self.set_content_in_language(alert, 'zh-hant', 'Traditional Chinese description')
+        alert_from_db = validate_save_and_reload(alert)
+
+        url = '/v1/content/alerts/zh_TW/{0}/'.format(alert_from_db.pk)
```
rasmus-storjohann-PG, line 89: There should be tests for both client side locale codes.

content/viewsets.py
```diff
@@ -2,6 +2,7 @@
 from rest_framework import viewsets
 from common.filters import (AlertIdFilter)
 from content import models, serializers, documentation
+from content.convert_locale_code import (convert_locale_code)
```
rasmus-storjohann-PG, line 5: `()` are not needed here, they are useful when you import so many symbols from one package that they don't fit on a line

content/tests/test_api.py
tomy-pg: I agree, but I am unsure how this would look like. Hmm.. perhaps we can chat about this more if the solution is straight forward or shelf this for now so this can make release. I thought the `AlertBuilder` was designed to build only one content alert with a language code, header and content. To be able to do what is suggested above, it seems like the`AlertBuilder` would have to be able to build one or many content alerts.

rasmus-storjohann-PG: This is for a future round of this, since we want to get this in. The way to do this is to write the client code first, making the API easy and natural to use, and then write the implementation of the builder afterwards to do what's needed, including whatever internal data structures the builder might need. If you want to try it out in a second PR against this same issue, you can do that.

PR #565, opened 2020-10-14
Feature/issue 557 import bc211 csv

rasmus-storjohann-PG:


rasmus-storjohann-PG:
Add test to cover unicode characters.

bc211/csv_import/parser.py
```diff
@@ -0,0 +1,242 @@
...
+    if header == 'TaxonomyCodes':
+        taxonomy_terms += parse_taxonomy_terms(value, 'AIRS')
+
+
+def parse_taxonomy_terms(value, vocabulary):
+    if vocabulary == 'AIRS':
+        names = re.split(r'[;\* ]', value)
+    else:
+        names = re.split(r'[;\-\* ]', value)
+    return [build_taxonomy_object(name, vocabulary) for name in names if name]
```
tomy-pg, line 152: From QA @ian-lin-PG and I are seeing in the taxonomy `name` column that taxonomies that should be together are separated into different rows. 
See screenshot below. 
Screen Shot 2020-10-26 at 4 17 40 PM

tomy-pg (changes requested):

bc211/csv_import/parser.py
```diff
@@ -54,7 +55,17 @@ def parse(sink, lines):
 def parse_organization_and_service_fields(header, value, organization_or_service):
     output_header = organization_header_map.get(header, None)
     if output_header:
-        organization_or_service[output_header] = value
+        if output_header == 'last_verified_on':
+            organization_or_service[output_header] = fix_date_time_string(value)
```
tomy-pg, line 63: If the `last_verified_on` is empty, shouldn't we return empty as opposed to returning the date of when the script was run?

bc211/csv_import/parser.py
```diff
@@ -36,6 +36,9 @@ def parse(sink, lines):
             if header == 'ParentAgencyNum':
                 parent_id = value
 
+        if not organization_or_service['id']:
+            organization_or_service['id'] = str(uuid.uuid4())
```
ian-lin-PG: I don't think we should generated service or organization ids given that BC211 are responsible for creating the ids. I think a better approach is to log the services/organizations without ids and forward them to BC211. As @tomy-pg pointed out, if a bookmarked service has its id changed we can dealing with complication down the road.

bc211/csv_import/parser.py
```diff
@@ -37,7 +37,8 @@ def parse(sink, lines, vocabulary=None):
                 parent_id = value
 
         if not organization_or_service['id']:
-            organization_or_service['id'] = str(uuid.uuid4())
+            organization_or_service['id'] = compute_hash(organization_or_service['name'],
+                                                         organization_or_service['alternate_name'])
```
tomy-pg: I like this solution better, but we can run into errors if bc211 eventually adds an id to the organization / service and a user has one of these bookmarked on their app. It may be that we want to log out then report these organizations and services that have missing ids to bc211 before we add them to the processed csv.

bc211/csv_import/parser.py
```diff
@@ -66,6 +66,8 @@ def parse_organization_and_service_fields(header, value, organization_or_service
 
 
 def fix_date_time_string(date_time_string):
+    if not date_time_string:
+        return None
```
tomy-pg, line 70: Oh nice. Omit previous comment on this.

rasmus-storjohann-PG: Should rename the function to make this not surprising

bc211/management/commands/import_bc211_csv_data.py
```diff
@@ -0,0 +1,22 @@
+import argparse
```
rasmus-storjohann-PG, line 1: rename to `convert_bc211_csv_to_openreferral_csv`

bc211/csv_import/parser.py
```diff
@@ -0,0 +1,306 @@
...
+    hasher = hashlib.sha1()
+    for arg in args:
+        hasher.update(arg.encode('utf-8'))
+    return hasher.hexdigest()
+
+
+def set_id_if_not_already_set(organization_or_service):
+    if not organization_or_service['id']:
+        organization_or_service['id'] = compute_hash(organization_or_service['name'],
+                                                     organization_or_service['alternate_name'])
```
rasmus-storjohann-PG, line 209: Terminate with an error on this condition

bc211/csv_import/parser.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+def parse_phone_number_fields(header, value, phone_numbers):
+    output_phone_header = phone_header_map.get(phone_header_with_index_one(header), None)
+    if output_phone_header:
+        phone_index = get_zero_based_phone_index(header)
+        set_phone_array_length(phone_numbers, phone_index)
+        phone_numbers[phone_index][output_phone_header] = value
+        if header == 'PhoneFax':
+            phone_numbers[phone_index]['type'] = 'Fax'
```
tomy-pg, line 150: The `phones.csv` generated only contains phone types that are `Fax`. The rest of the rows do not contain a type. 

Looking at the iCarol.csv: `PhoneType1` does not seem to have any values inside of its cells. For this, we could possibly do something like:

```
if header == PhoneType1:
     phone_numbers[phone_index]['type'] = 'PhoneNumber1' # or Phone1, Primary phone
```

tomy-pg (changes requested):

tomy-pg (approved):

PR #570, opened 2020-10-27
Feature/issue 549 support multiple api keys

rasmus-storjohann-PG:


tomy-pg (approved):
Looks good to me. Some notes: 
- Followed your steps for QA. 
- I was unable to properly post / put a push notification using Postman to see if this works. I kept receiving a `404`. Instead I used the client to test that these changes work by swapping the API key from valid / invalid to see if they would be saved on the PR's Heroku DB.

PR #573, opened 2020-11-17
Feature/issue 564 import open referral csv

tomy-pg:


bc211/open_referral_csv_import/parser.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def parse_organization_id(value):
+    organization_id = parse_required_field('organization_id', value)
+    return remove_double_escaped_html_markup(organization_id)
+
+
+def parse_service_id(value):
+    service_id = parse_required_field('service_id', value)
+    return remove_double_escaped_html_markup(service_id)
```
rasmus-storjohann-PG, line 14: In those cases where these functions are identical in terms of their behaviour, I think you'll find that combining them and naming them to reflect what they do (parse_required_field_with_double_escaped_html()) will make the code at the calling site more readable.

rasmus-storjohann-PG (changes requested):

bc211/open_referral_csv_import/is_inactive.py
```diff
@@ -0,0 +1,3 @@
+def is_inactive(description):
+    # This is BC211's convention for marking records as inactive
+    return description and description.strip().startswith('DEL')
```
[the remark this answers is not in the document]
tomy-pg: I have refactored `is_inactive` to work with both xml and open referral csv files. While importing the open referral csv files, I noticed some inactive records were not `DEL` but varied in casing such as `Del` so I made this case insensitive.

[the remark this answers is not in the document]
tomy-pg: I noticed most inactive records also included a number directly after `DEL`, `DEL16`.
In one case, I also noticed an inactive record that contained a whitespace character between DEL and a number, `DEL 16`. I have included these cases in the function to improve it and also make it less susceptible to errors since I make the `DEL` look-up case insensitive.

bc211/open_referral_csv_import/service_taxonomy.py
```diff
@@ -0,0 +1,40 @@
...
+            raise
+
+
+expected_headers = ['id', 'service_id', 'taxonomy_id', 'taxonomy_detail']
+
+
+def import_service_taxonomy(row):
+    service_id = parser.parse_service_id(row[1])
+    taxonomy_id = parser.parse_taxonomy_id(row[2])
+    save_service_taxonomy_term(service_id, taxonomy_id)
```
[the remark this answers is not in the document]
tomy-pg: The TaxonomyTerm model services_taxonomy_terms is built from does not have the `id` column defined so Django creates the id for us. Let me know if you want this behaviour to change.

bc211/open_referral_csv_import/address.py
```diff
@@ -0,0 +1,64 @@
...
+
+def build_address_active_record(row):
+    active_record = Address()
+    active_record.city = parser.parse_city(row[8])
+    active_record.country = parser.parse_country(row[12])
+    active_record.attention = parser.parse_attention(row[3])
+    active_record.address = parser.parse_address(row[4])
+    active_record.state_province = parser.parse_state_province(row[10])
+    active_record.postal_code = parser.parse_postal_code(row[11])
+    return active_record
```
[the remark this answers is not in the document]
tomy-pg: The `Address` model that we use does not have the `id` column defined so Django auto-generates the id for us. I recognize in your work you generate an id as well. Do you want me to change this or leave the model as is?

I will make the change to handle all `address_x` fields.

bc211/open_referral_csv_import/parser.py
tomy-pg: Are you suggesting that instead of:
`active_record.organization_id = parse_organize_id(...)`
`active_record.name = parse_name(...)`
to do:
`active_record.organization_id = parse_required_field_with_double_escaped_html(...)`
`active_record.name = parse_required_field_with_double_escaped_html(...)`
?

bc211/open_referral_csv_import/address.py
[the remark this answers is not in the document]
rasmus-storjohann-PG: Change it so that it uses the ids I generate in my PR. It's so that repeated imports of the same data will give consistent ids each time.

bc211/open_referral_csv_import/tests/test_sandbox.py
```diff
@@ -0,0 +1,42 @@
...
+        address_3_id = a_string()
+        self.address_1 = AddressBuilder().with_id(address_1_id).with_city('Richmond').create()
+        self.address_2 = AddressBuilder().with_id(address_2_id).with_city('Vancouver').create()
+        self.address_3 = AddressBuilder().with_id(address_3_id).with_city(None).create()
+
+        self.physical_address_type = AddressType.objects.get(pk='physical_address')
+
+        LocationAddress(address=self.address_1, location=self.location_1, address_type=self.physical_address_type).save()
+        LocationAddress(address=self.address_2, location=self.location_2, address_type=self.physical_address_type).save()
+        LocationAddress(address=self.address_3, location=self.location_3, address_type=self.physical_address_type).save()
```
rasmus-storjohann-PG, line 33: I would have thought that this would work
```py
LocationAddress(address=self.address_3, location=self.location_3, address_type='physical_address').save()
```

bc211/open_referral_csv_import/phone.py
```diff
@@ -0,0 +1,76 @@
...
+    except ObjectDoesNotExist as error:
+        pass
+    except CsvParseException:
+        pass
+
+
+def build_phone_number_type_active_record(row):
+    active_record = PhoneNumberType()
+    active_record.id = parser.parse_required_type(row[8])
+    return active_record
```
rasmus-storjohann-PG, line 67: The naming convention is that "build" means make the AR instance, and "create" means make the AR and save it to the DB. So you might want to include the "save" step in the helper function and rename it to "create_..."

bc211/open_referral_csv_import/service_at_location.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def read_and_import_rows(reader, collector, counters):
+    for row in reader:
+        if not row or service_at_location_has_invalid_data(row, collector):
+            continue
+        import_service_at_location(row, counters)
+
+
+def service_at_location_has_invalid_data(row, collector):
```
rasmus-storjohann-PG, line 39: Inconsistent naming, inactive, not invalid

human_services/locations/models.py
```diff
@@ -18,6 +18,7 @@ class Location(ValidateOnSaveMixin, TranslatableModel):
     point = models.PointField(blank=True, null=True)
     translations = TranslatedFields(
         name=models.CharField(max_length=200),
+        alternate_name=models.CharField(blank=True, max_length=200),
```
rasmus-storjohann-PG, line 21: Django's handling of fields where blank and null are different is messy, so we want to consistently set both to true or both to false. I also don't want to remember what their values default to, so they should always both be given regardless.

human_services/addresses/tests/test_models.py
```diff
@@ -18,14 +18,23 @@ def test_empty_address_field_saves_as_null(self):
...
         address = AddressBuilder().build()
         address_from_db = validate_save_and_reload(address)
         self.assertEqual(address_from_db.city, address.city)
 
-    def test_city_field_cannot_be_empty(self):
-        with self.assertRaises(exceptions.ValidationError):
-            AddressBuilder().with_city('').create()
+    def test_empy_city_field_saves_as_null(self):
+        address = AddressBuilder().with_city('').create()
+        self.assertTrue(Address.objects.first().city is None)
```
rasmus-storjohann-PG, line 37: nice

bc211/tests/test_importer.py
```diff
@@ -151,7 +151,7 @@ def test_do_not_import_inactive_location(self):
     def test_do_not_import_inactive_service(self):
         organization = OrganizationBuilder().create()
         location = LocationBuilder(organization).create()
-        inactive_description = 'DEL ' + a_string()
+        inactive_description = 'DEL17 ' + a_string()
```
rasmus-storjohann-PG, line 154: What is the significance of the two digits? What is the expected behaviour if the description starts with 'DEL78'? I don't think the digits mean anything, in which case they should not appear in the test.

bc211/service.py
```diff
@@ -37,7 +37,7 @@ def update_services_for_location(location_id, services, counters):
 
 
 def save_service_if_needed(service, counters):
-    if is_inactive(service):
+    if is_inactive(service.description):
```
rasmus-storjohann-PG, line 40: I like the old way better, because the fact that the description field is where the DEL is found is delegated to the helper function ... with the new code, the higher level code needs to include that detail, which is not as clean for separation of concerns. But if you have good reason for this change, I'm open to it.

bc211/open_referral_csv_import/tests/test_parser.py
```diff
@@ -0,0 +1,102 @@
...
+        self.assertEqual(f'{address_1}\n{address_2}\n{address_3}\n{address_4}', parsed_addresses)
+
+    def test_can_parse_addresses_when_some_fields_are_empty(self):
+        address_1 = a_string()
+        address_2 = a_string()
+        address_3 = ''
+        address_4 = ''
+        addresses = [address_1, address_2, address_3, address_4]
+        parsed_addresses = parser.parse_addresses(addresses)
+        self.assertEqual(f'{address_1}\n{address_2}', parsed_addresses)
```
rasmus-storjohann-PG, line 43: nice

rasmus-storjohann-PG (changes requested):

bc211/is_inactive.py
```diff
@@ -1,3 +1,11 @@
...
+
+# BC211's convention for marking records as inactive is DEL + number or DEL + whitespace + number
+
+def is_inactive(description):
+    return description and description_starts_with_inactive_record_convention(description)
+
+
+def description_starts_with_inactive_record_convention(description):
+    stripped_description = description.strip()
+    return re.search(r'^[D-d-E-e-L-l]{3}[0-9\s]', stripped_description)
```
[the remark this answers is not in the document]
tomy-pg: There is already a test case that exists which ensure we don't accidentally mark strings starting with "Del..." as inactive.

bc211/open_referral_csv_import/address.py
```diff
@@ -0,0 +1,86 @@
...
+        counters.count_address()
+        location_address_active_record = build_location_address_active_record(address_active_record, row, collector)
+        location_address_active_record.save()
+        counters.count_location_address()
+    except ValidationError as error:
+        LOGGER.warn('{}'.format(error.__str__()))
+    except CsvParseException:
+        pass
+    except ObjectDoesNotExist:
+        pass
```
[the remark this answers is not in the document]
tomy-pg: These will be handled or removed once your PR is in. I have these here temporarily to be able to test the script.

bc211/service.py
tomy-pg: Previously we were building active records, then checking to see if the record is inactive.

I made this change so that I could first parse the description, see if BC211 marked the record as inactive and then build the record. 

I am open to changing this back if you would like.

bc211/open_referral_csv_import/service_taxonomy.py
```diff
@@ -0,0 +1,98 @@
...
+            taxonomy_updates_for_service.append(last_service)
+    except ValidationError as error:
+        LOGGER.warning('%s', error.__str__())
+    except ObjectDoesNotExist as error:
+        pass
+
+
+def bulk_update_taxonomy_updates_for_service(taxonomy_updates_for_service):
+    if taxonomy_updates_for_service:
+        Service.objects.bulk_update(taxonomy_updates_for_service)
```
tomy-pg, line 76: I am using `bulk_update ` here as opposed to `bulk_create` because the `ServiceTaxonomyTerm` intermediary table is not explicitly declared. Therefore, I cannot use the ORM to do something like ServiceTaxonomyTerm.objects.bulk_create()
We seem to create the table like so:
```
/human_services/services/models
class Service():
    id = ...
    organization = ...
    taxonomy_terms = models.ManyToManyField(TaxonomyTerm,
                                            db_table='services_service_taxonomy_terms')
```
As part of this PR, I can create the ServiceTaxonomyTerm model explicitly. What are your thoughts on this @rasmus-storjohann-PG ?

bc211/open_referral_csv_import/tests/test_sandbox.py
tomy-pg: `address_type` expects an `AddressType` instance

rasmus-storjohann-PG (changes requested):

bc211/open_referral_csv_import/tests/test_importer.py
```diff
@@ -0,0 +1,473 @@
...
+        the_latitude = a_latitude_as_a_string()
+        the_longitude = a_longitude_as_a_string()
+        location_data = (OpenReferralCsvLocationBuilder(self.organization).
+                        with_latitude(the_latitude).
+                        with_longitude(the_longitude).
+                        build())
+        import_location(location_data, InactiveRecordsCollector(), ImportCounters())
+        locations = Location.objects.all()
+        self.assertEqual(locations[0].point.x, float(the_longitude))
+        self.assertEqual(locations[0].point.y, float(the_latitude))
```
rasmus-storjohann-PG, line 210: Conversion from float to string always succeeds, conversion from string to float can fail. So when you have the choice, it's better to use the float -> string form. Here you could have done `the_latitude = a_latitude()`, `with_latitude(str(the_latitude))` and `self.assertEqual(locations[0].point.x, the_longitude)`

bc211/open_referral_csv_import/tests/test_importer.py
```diff
@@ -0,0 +1,473 @@
...
+        self.assertEqual(services_at_location[0].service_id, self.service_id)
+
+    def test_can_import_location_id(self):
+        service_data = OpenReferralCsvServiceAtLocationBuilder(self.service, self.location).build()
+        import_service_at_location(service_data, ImportCounters())
+        services_at_location = ServiceAtLocation.objects.all()
+        self.assertEqual(services_at_location[0].location_id, self.location_id)
+
+
+class OpenReferralAddressImporterTests(TestCase):
```
rasmus-storjohann-PG, line 234: You can drop the `OpenReferral` prefix here, you hare not using it consistently and it hurts readability

bc211/open_referral_csv_import/tests/test_importer.py
```diff
@@ -0,0 +1,473 @@
...
+
+    def test_can_import_city(self):
+        the_city = a_string()
+        address_data = OpenReferralCsvAddressBuilder(self.location).with_city(the_city).build()
+        import_address(address_data, ImportCounters())
+        addresses = Address.objects.all()
+        self.assertEqual(addresses[0].city, the_city)
+
+    def test_can_import_country(self):
+        the_country = a_string(2, string.ascii_uppercase)
```
rasmus-storjohann-PG, line 256: See the established pattern for `a_website_address()` and `an_email_address()`, you should create `a_country_code()`.

bc211/open_referral_csv_import/tests/test_importer.py
```diff
@@ -0,0 +1,473 @@
...
+        import_address(address_data, ImportCounters())
+        addresses = Address.objects.all()
+        self.assertEqual(addresses[0].address, the_address_1)
+
+    def test_can_import_addresses(self):
+        the_address_1 = a_string()
+        the_address_2 = a_string()
+        address_data = (OpenReferralCsvAddressBuilder(self.location).
+                            with_address_1(the_address_1).
+                            with_address_2(the_address_2).
```
rasmus-storjohann-PG, line 287: You need to test all four fields, this is a typical area where bugs can hide for a long time, since most addresses don't use all four lines.

bc211/open_referral_csv_import/tests/test_importer.py
```diff
@@ -0,0 +1,473 @@
...
+        self.location = LocationBuilder(organization).with_id(self.location_id).create()
+
+    def test_can_import_phone_number_type(self):
+        the_phone_type = a_string()
+        phone_data = (OpenReferralCsvPhoneBuilder(self.location).
+                        with_phone_type(the_phone_type).
+                        build())
+        import_phone(phone_data, InactiveRecordsCollector(), ImportCounters())
+        phone_number_types = PhoneNumberType.objects.all()
+        self.assertEqual(phone_number_types[0].id, the_phone_type)
```
rasmus-storjohann-PG, line 368: I don't see why this is its own test suite, can be combined with the test suite below

rasmus-storjohann-PG (changes requested):

rasmus-storjohann-PG:
OK, so I've read through the tests and they look good, see a few comments above. Some functions that are complicated enough that they should have tests: `service_taxonomy#read_and_import_rows()`,

rasmus-storjohann-PG (changes requested):

human_services/locations/models.py
```diff
@@ -49,8 +49,6 @@ class LocationAddress(ValidateOnSaveMixin, models.Model):
                                  related_name='location_addresses')
     address_type = models.ForeignKey(AddressType, on_delete=models.PROTECT)
 
-    class Meta:
-        unique_together = ('location', 'address_type')
```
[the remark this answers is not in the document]
tomy-pg: From our discussion, we noticed that there are seemingly repeated rows of the same location and address type. I have since brought this `unique_together` constraint and will log out the validation errors with information of the row violating it.

Issue #617, opened 2021-03-12
When computing related tasks, ensure that BC tasks and MB tasks are kept separate


tomy-pg:
See #615

### Acceptance Criteria 
- [x]  Compute related tasks using NLP. Ensure that no BC task is related to a MB task and vice versa.

### Steps for QA: 
1.

ian-lin-PG:
Randome checks pass QA 1.66.0

PR #624, opened 2021-03-29
Issue #617 Create a separate command line tool to remove stale recomm…

rasmus-storjohann-PG:
…endation data

### DIFF
--- a/utility/prepare_deploy.sh
+++ b/utility/prepare_deploy.sh
@@ -205,6 +205,10 @@ then
     checkForSuccess "convert Manitoba WIN data, output to $NewcomersGuidePath"
 fi
 
+echo "removing existing similarity scores"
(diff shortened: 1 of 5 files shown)


utility/prepare_deploy.sh
```diff
@@ -205,6 +205,10 @@ then
     checkForSuccess "convert Manitoba WIN data, output to $NewcomersGuidePath"
 fi
 
+echo "removing existing similarity scores"
+./manage.py remove_all_topics_and_recommendations
+checkForSuccess "remove exising similarity scores"
```
rasmus-storjohann-PG, line 210: So this code (1) maintains the same functionality as before, but (2) doesn't probably do anything in the context of this script, since these tables are empty at this point anyway.

PR #625, opened 2021-03-29
Feature/issue 619 region specific service search

tomy-pg:


common/testhelpers/random_test_values.py
```diff
@@ -69,4 +69,4 @@ def a_country_code():
     return a_string(2, string.ascii_uppercase)
 
 def a_region_specific_id(region):
-    return '{0}_{1}'.format(a_string(), region)
\ No newline at end of file
+    return '{0}_{1}'.format(a_string(), region.lower())
```
rasmus-storjohann-PG, line 72: This is the wrong place to convert to lower case. What if I wanted to test the behaviour when an upper case region is used, this generator would not work for me and it would be unclear why

common/filters.py
```diff
@@ -194,6 +194,6 @@ def filter_queryset(self, request, queryset, view):
         region = request.query_params.get('region', None)
         if region:
             if queryset.model is ServiceAtLocation:
-                service_id_suffix = '_' + region
+                service_id_suffix = '_{0}'.format(region.lower())
```
rasmus-storjohann-PG, line 197: There needs to be a test for this

human_services/services_at_location/tests/test_api.py
```diff
@@ -501,4 +502,31 @@ def test_can_return_id(self):
         service_at_location = ServiceAtLocationBuilder().create()
         response = self.client.get('/v1/services_at_location/')
         json = response.json()
-        self.assertEqual(json[0]['id'], service_at_location.id)
\ No newline at end of file
+        self.assertEqual(json[0]['id'], service_at_location.id)
+    
+    def test_can_filter_by_region(self):
+        client_region = 'mb'
```
rasmus-storjohann-PG, line 508: remove this variable and just use `'mb'`, I think you will find that the test is easier to read

rasmus-storjohann-PG (changes requested):
Nice work, just a couple of suggestions

common/filters.py
rasmus-storjohann-PG: Just a test that shows what happens when you pass in `'BC'` as your region

rasmus-storjohann-PG: The test should go next to `test_can_filter_by_region()`

PR #635, opened 2021-05-04
Feature/issue 634 fix deployment

rasmus-storjohann-PG:


utility/prepare_deploy.sh
```diff
@@ -220,9 +220,13 @@ if [ "$mb211Path" != "" ]
...
 checkForSuccess "compute similarity scores"
 
-echo "adding manual similarity scores ..."
+echo "adding manual similarity scores for bc..."
 ./manage.py manage_manual_recommendations $ManualRecommendations --region bc
-checkForSuccess "add manual similarity scores"
+checkForSuccess "add manual similarity scores for bc"
+
+echo "adding manual similarity scores for mb..."
+./manage.py manage_manual_recommendations "${ManualRecommendations}mb/"
```
rasmus-storjohann-PG, line 228: The shell will accept `//` as equivalent to `/`, so I think it's safer to make this `.../mb/` so it'll work whether or not ManualRecommendations has a trailing `/`. Defensive programming

PR #636, opened 2021-05-06
WorkInProgress Feature/issue 633 csv to algolia json

rasmus-storjohann-PG:


rasmus-storjohann-PG:
This approach has been abandoned, using mongo instead

PR #671, opened 2021-11-10
Feature/issue 670 fix push notification failures

brandon-PG:


rasmus-storjohann-PG:
I made one commit that makes the logic a little easier to follow, specifically answers: How long will the retry logic wait (about 10 secs) and added some brackets to make a calculation with `*` and the rarely used `**` unambiguous.

